PR #42, opened 2026-04-10
feature/SAV-5436 stepper component

JenLoVangas:


tomas562:
Stepper component must follow MUI API and wrap existing stepper component

PR #43, opened 2026-04-13
SAV-5460 feat: refactor icon components to use Phosphor icons, add new icons

JenLoVangas:
Refactor icon components to use Phosphor icons, add new icons (Barcode, CircleFilled, DotCircleFilled, MapPinArea, Scales), and improve overall consistency with a unified Storybook showcase


src/assets/icons/CheckBoldDisabledIcon.tsx
tomas562, line 1: this is identical icon to src/assets/icons/CheckBoldIcon.tsx. Do we need it? Consumers will silently loose distinction because they look the same

src/library/icons.ts
tomas562, line 1: not all icons are exported

PR #44, opened 2026-04-13
SAV-5458 fix: add iconOnly prop to RcSesButton and update button styles

JenLoVangas:


src/components/common/Button/index.tsx
```diff
@@ -1,18 +1,48 @@
...
+      {...rest}
+      size={size}
+      variant={variant}
+      sx={[
+        ...(Array.isArray(sx) ? sx : [sx]),
+        isIconOnly
+          ? {
+              minWidth: 0,
+              padding: 0,
+              width: iconOnlySizingBySize[size],
```
tomas562: You should set height: iconOnlySizingBySize[size], as well because in rc-ses-react-components\src\theme\light\MuiButton.tsx
```
 styleOverrides: {
    root: {
      ...
      height: '2.75rem',
```
and icons won't be square for all sizes

src/components/common/Button/index.tsx
```diff
@@ -1,18 +1,48 @@
...
 
 function RcSesButton(props: Props) {
-  const { children, ...rest } = props
+  const { children, iconOnly, size = 'medium', sx, variant, ...rest } = props
+
+  const currentVariant = variant ?? defaultProps.variant
+  const isIconOnly =
+    !!iconOnly && (currentVariant === 'contained' || currentVariant === 'outlined')
+
+  const iconOnlySizingBySize = {
```
tomas562: iconOnlySizingBySize is in render function. That's a bad practice because map is recreated on every render. It should be a module-level constant. Also such constants should be named in capital letters. My name suggestion: 
`
const ICON_ONLY_SIZE_MAP = {
  small: '2.25rem',
  medium: '2.75rem',
  large: '3rem',
} as const
`

tomas562:
image
disabled button does not follow figma colors:
image

src/components/common/Button/index.tsx
JenLoVangas: pataisyta

src/components/common/Button/index.tsx
JenLoVangas: pataisyta

PR #50, opened 2026-04-16
SAV-5430 feat: Card and stepper connection

TautvydasBliudzius:


src/components/layout/ServiceWizardStepper/ServiceWizardStepper.css
```diff
@@ -1,4 +1,4 @@
-.ServiceWizardStepper {
+.ServiceWizardStepper .ServiceWizardStepper-horizontal {
```
tomas562: descendant selector, but both classes are on the SAME element. this CSS selector will never match

src/components/layout/ServiceFormContainer/CardFormContainer/index.tsx
```diff
@@ -0,0 +1,50 @@
+import { Container } from '@mui/system'
```
tomas562: import { Container } from '@mui/material'

src/library/index.ts
```diff
@@ -67,10 +69,12 @@ export { RcSesSelect }
 export { RcSesTab, RcSesTabPanel, RcSesTabs, RcSesTabsWrapper }
 export { RcSesTextField }
 export { DataPagination }
+export { ServiceWizardStepper }
 
 export {
   RcSesServiceFormActions,
-  RcSesServiceFormContainer,
```
tomas562, line 73: removing RcSesServiceFormContainer will be breaking change for consumers this component

src/components/layout/ServiceFormContainer/CardFormContainer/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+        },
+        gap: '24px',
+        paddingBottom: '64px',
+      }}
+    >
+      {steps.length > 2 && (
+        <ServiceWizardStepper
+          steps={steps}
+          onStepClick={onStepClick}
+          orientation={getLayoutDirection() === 'row' ? 'vertical' : 'horizontal'}
```
tomas562: Better practice would be set constant and use it's values instead of invoking function every time 
`const layoutDirection = getLayoutDirection()`

src/components/layout/ServiceFormContainer/CardFormContainer/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { StepItem } from '../../ServiceWizardStepper/StepperTypes'
+
+export type LayoutMode = 'column' | 'row'
+
+export interface RcSesCardFormContainerProps extends RcSesCardProps {
+  steps: StepItem[]
+  onStepClick?: (steps: StepItem[]) => void
+  layout?: LayoutMode
+  isLoading?: boolean
+}
```
tomas562, line 15: Export types from library. Consumers might want to use these types as well in there code

PR #55, opened 2026-04-21
SAV-5450 feat: List with icons

TautvydasBliudzius:


src/components/common/ListWithIcons/components/ListWithIconsItem.tsx
```diff
@@ -0,0 +1,82 @@
...
+      >
+        <Box
+          sx={{
+            display: 'inline-flex',
+            alignItems: 'center',
+            gap: 0.5,
+            maxWidth: { xs: '100%', md: 'none' },
+          }}
+        >
+          <Box sx={{ display: 'flex', alignItems: 'center', width: 20 }}>{icon}</Box>
```
tomas562: icon is optional in type but <Box> is rendered always.

src/components/common/ListWithIcons/components/ListWithIconsItem.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+const ListWithIconsItem = ({ icon, text, disabled, tooltip, isLoading }: Props) => (
+  <Box
+    component='li'
+    sx={{
+      display: 'flex',
+      alignItems: 'center',
+      width: { xs: '100%', md: isLoading ? '120px' : 'auto' },
+      gap: 0.5,
+      height: 20,
```
tomas562: hardcoded height looks suspicious to me. if content will not fit into that height ugly things might happen..

src/components/common/ListWithIcons/components/ListWithIconsItem.tsx
```diff
@@ -0,0 +1,82 @@
...
+        variant='rectangular'
+        sx={{
+          flex: 1,
+          height: 16,
+          borderRadius: 1,
+          width: { xs: '100%', md: 120 },
+        }}
+      />
+    ) : (
+      <Tooltip
```
tomas562: tooltip block will be rendered even if no tooltip provided. Maybe we should think about conditionally rendering it?

src/components/common/DataPagination/index.tsx
tomas562, line 17: I just noticed that pagination does not have loading state. Maybe we should consider adding it?

tomas562: will be added with other task 5451

PR #64, opened 2026-04-23
SAV-5451 feat: Selectable card list

TautvydasBliudzius:
## 🔗 Task

[SAV-5451](https://jira.registrucentras.lt/jira/browse/SAV-5451)

## 🧾 Summary

This task delivered end-to-end support for the new list/card selection components across implementation quality, tests, and Storybook.

## 📸 Screenshots

image

## ✅ Checklist

 [+] Component added/updated in Storybook (if applicable)
 [+] Tests added/updated
 [+] UI matches approved design (Figma)
 [+] Accessibility reviewed (keyboard navigation, labels, semantics)

## ⚠️ Risks

1. Story examples depend on current item shape/types; future type changes in list item models may require story/test updates.
2. Selectable card interaction currently relies on click behavior and radio visual state; keyboard/a11y behavior should be explicitly verified.
3. Pagination behavior is tested for common paths, but edge-case behavior with dynamic data replacement should be monitored in integration.


src/components/common/DataPagination/components/DataPaginationSkeleton.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const itemSize = 40
+  return (
+    <>
+      <Skeleton variant='rounded' width={buttonWidth} height={itemSize} />
+      <Box
+        sx={{
+          display: 'flex',
+          gap: '6px',
+          px: '3px',
+          borderRadius: '8px',
```
saukaite: 1. If we use hardcoded values here, I suppose we don't have any tokens, right? Probably out of scope for this task, but would be nice to have some tokens. Then we can use those tokens where needed.
2. I see we use both `rem` and `px`, probably would make sense to keep one option for consistency? Any agreements here?

tomas562:
@TautvydasBliudzius please solve conflicts before we can start review

src/components/common/SelectableCardList/components/SelectableCardListItem.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+      <SelectableCardListItem
+        title='Main title'
+        subtitle='Some subtitle'
+        selected={false}
+        onSelect={vi.fn()}
+      />,
+    )
+
+    expect(screen.getByText('Main title')).toBeInTheDocument()
+    expect(screen.getByText('Some subtitle')).toBeInTheDocument()
```
saukaite: The test would still pass if we swapped text, maybe we can change to smth like:

`    expect(screen.getByRole('heading', { name: 'Main title' })).toBeInTheDocument()
`

src/components/common/ListWithIcons/index.tsx
```diff
@@ -1,47 +1,55 @@
...
-  icon?: React.ReactNode
-  text?: string
-  disabled?: boolean
-  tooltip?: string
-}
+import ListWithIconsItem, { ListWithIconsItemData } from './components/ListWithIconsItem'
 
 export interface ListWithIconsProps {
-  items: ListWithIconsItemData[]
+  items?: ListWithIconsItemData[]
```
saukaite: Are `items` optional intentionally? Now all props are optional, I think each component should have required props so it can function correctly.

src/components/common/SelectableCardList/index.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+    id: `item-${i + 1}`,
+    title: `Item ${i + 1}`,
+    subtitle: `Subtitle ${i + 1}`,
+  }))
+
+describe('SelectableCardList', () => {
+  test('renders skeleton cards when items is undefined', () => {
+    const { container } = renderList(<SelectableCardList onSelect={vi.fn()} />)
+
+    expect(container.querySelectorAll('.MuiSkeleton-root').length).toBeGreaterThan(0)
```
saukaite: If the library changes its class name, the test would break. I'd prefer to get by role or test id.

src/examples/ListWithPagination/index.tsx
```diff
@@ -1,72 +1,114 @@
...
+      },
+      {
+        id: 'address',
+        icon: MapPinAreaIcon,
+        text: item.address,
+      },
+    ],
+  }))
 
 function ListWithPagination() {
```
saukaite, line 55: Just a question - why we need such component in the `examples`? What's the purpose?

src/components/common/SelectableCardList/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+    <Container
+      sx={{
+        display: 'flex',
+        flexDirection: 'column',
+        alignItems: 'center',
+        gap: '16px',
+        paddingBottom: '64px',
+      }}
+    >
+      {error || (
```
saukaite: `error` is a `ReactNode`, so it may be falsy. Using || can skip valid values.
I'd prefer `{error ? error : (...)}` or `{error != null ? error : (...)}` patterns in this case.

saukaite (commented):
Left a few comments/questions, feel free to ping once ready for re-review 🙌

src/examples/ListWithPagination/index.tsx
TautvydasBliudzius: The purpose of this example is to provide a concrete usage scenario for the component. It helps validate integration (pagination + selection + loading states) outside of isolated unit tests and makes it easier to manually verify behavior during development.

saukaite: Thanks for the explanation, @TautvydasBliudzius 👍

src/components/common/SelectableCardList/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+    ))
+
+  return (
+    <Container
+      sx={{
+        display: 'flex',
+        flexDirection: 'column',
+        alignItems: 'center',
+        gap: '16px',
+        paddingBottom: '64px',
```
JenLoVangas: Replace paddingBottom: '64px' with [padding: '0px'] so that the component has no internal spacing and can fill the entire parent element's area.

src/components/common/SelectableCardList/components/SelectableCardListItem.tsx
```diff
@@ -0,0 +1,108 @@
...
+    <Stack direction='row' alignItems='flex-start' gap='0.75rem'>
+      <Radio
+        checked={selected}
+        sx={{
+          p: 0,
+          height: '1.25rem',
+          width: '1.25rem',
+          margin: '0.125rem 0rem',
+        }}
+      />
```
tomas562, line 42: radio has no aria-label. Let's keep UI friendly for all device types

src/components/common/ListWithIcons/index.tsx
```diff
@@ -1,46 +1,50 @@
...
 
-import ListWithIconsItem from './components/ListWithIconsItem'
-
-export interface ListWithIconsItemData {
-  id: string
-  icon?: React.ReactNode
-  text?: string
-  disabled?: boolean
-  tooltip?: string
-}
```
tomas562, line 11: Interface was moved and is now imported directly from the sub-component in three places. My Suggestion is to re-export it here and then update imports where needed `import { ListWithIconsItemData } from '@/components/common/ListWithIcons'`

src/components/common/ListWithIcons/index.tsx
```diff
@@ -1,46 +1,50 @@
...
+    {isLoading ? (
+      <>
+        {Array.from({ length: skeletonCount }).map((_, index) => (
+          // eslint-disable-next-line react/no-array-index-key
+          <ListWithIconsItem key={index} isLoading />
+        ))}
+      </>
+    ) : (
+      // eslint-disable-next-line react/no-array-index-key
+      items?.map((item, index) => <ListWithIconsItem key={index} {...item} />)
```
tomas562, line 47: items are not optional in ListWithIconsProps why we need .? here?

src/components/common/DataPagination/index.tsx
```diff
@@ -41,28 +44,48 @@ const DataPagination = ({
     onChange?.(newPage)
   }
 
+  const showSkeleton = isLoading && count === 0
+  const hidePagination = count <= 1 && !isLoading
```
tomas562: Conditions seem to have a gap. when isLoading = true and count =1 neither condition match. 
Either showSkeleton should be isLoading && count <= 1 or hidePagination should drop !isLoading guard..

src/components/common/SelectableCardList/index.tsx
```diff
@@ -0,0 +1,98 @@
+import { Box, Container } from '@mui/system'
```
tomas562: import { Box, Container } from '@mui/material'

tomas562 (approved):

PR #70, opened 2026-04-27
Fix: export ButtonProps type in library types

tomas562:
## 🔗 Task

[SAV-5458](https://jira.registrucentras.lt/jira/browse/SAV-5458)

## 🧾 Summary

`vite-plugin-dts` generates `.d.ts` files by traversing imports from the library entry point (`src/library/index.ts`). `ButtonProps` was only imported inside `src/components/common/Button/index.tsx` — not reachable from the entry — so the plugin never emitted `dist/types/buttons/ButtonProps.d.ts`. The generated `dist/components/common/Button/index.d.ts` referenced it via a relative path that didn't exist in `dist`, causing a `TS2307` error for consumers running `tsc`.

**Fix:** export `ButtonProps` from `src/library/types.ts`, making it reachable from the entry point so `vite-plugin-dts` emits the missing declaration file.

**Changed file:** `src/library/types.ts`

## 📸 Screenshots

N/A — type-only fix, no visual changes.

## ✅ Checklist

* [x] Tests added/updated

## ⚠️ Risks

Low. The change only adds a public type export — no runtime code is affected.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -16,18 +16,8 @@ module.exports = {
     'prettier',
     'plugin:storybook/recommended',
   ],
-  overrides: [
-    {
-      files: ['scripts/check-dist.ts'],
-      parserOptions: {
-        project: null,
-      },
-      rules: {
-        '@typescript-eslint/no-unused-vars': 'off',
-      },
-    },
-  ],
-  ignorePatterns: ['dist', '.eslintrc.cjs', 'vitest.config.ts'],
+  overrides: [],
+  ignorePatterns: ['dist', '.eslintrc.cjs', 'vitest.config.ts', 'scripts/check-dist.ts'],
   parser: '@typescript-eslint/parser',
   parserOptions: {
     project: './tsconfig.json',
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -16,7 +16,17 @@ module.exports = {
     'prettier',
     'plugin:storybook/recommended',
   ],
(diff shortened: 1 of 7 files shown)


saukaite (approved):
LGTM ✅

.eslintrc.cjs
```diff
@@ -17,7 +17,7 @@ module.exports = {
     'plugin:storybook/recommended',
   ],
   overrides: [],
-  ignorePatterns: ['dist', '.eslintrc.cjs', 'vitest.config.ts'],
+  ignorePatterns: ['dist', '.eslintrc.cjs', 'vitest.config.ts', 'scripts/check-dist.ts'],
```
saukaite, line 20: Does it cause issues if we run eslint for the `scripts/check-dist.ts`?

tomas562: yes. dist directory is created only after build. and It would check non existing path
image

tsconfig.check-dist.json
```diff
@@ -0,0 +1,18 @@
+{
+  "compilerOptions": {
+    "allowSyntheticDefaultImports": true,
+    "esModuleInterop": false,
```
saukaite, line 4:
```suggestion
    "esModuleInterop": true,
```
shouldn't it be `true`?

tomas562: config is aligned with the tsconfig.json. Also, since this is a type-only import import type {}, esModuleInterop doesn't affect anything here anyway.

PR #84, opened 2026-05-05
SAV-5592: add Snackbar component

saukaite:
## 🔗 Task

Task https://jira.registrucentras.lt/jira/browse/SAV-5592

🎨 [Figma designs for the component](https://www.figma.com/design/GPujqQEuxSUKDUPbYmx7TG/RC-Dizaino-sistema?node-id=45070-4564&t=dC1POzVmyP485SkC-0)

## 🧾 Summary

Added the new **Snackbar** component according to requirements.

## 📸 Screenshots

Added multiple cases in the Storybook, it's best to check there. Some examples:

Screenshot 2026-05-07 102425

Screenshot 2026-05-07 102435

Screenshot 2026-05-07 102519

## ✅ Checklist

* [x] Tests added/updated

## ⚠️ Risks

❓ Not sure if component animation is sufficient (might need to ask UX opinion)


src/components/common/Snackbar/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+  actionLabel,
+  open: controlledOpen,
+  duration = SNACKBAR_DURATION,
+  persist: controlledPersist,
+  index = 0,
+  showClose = true,
+}: RcSesSnackbarProps) {
+  const theme = useTheme()
+  const { t } = useTranslation('input')
+  const [internalOpen, setInternalOpen] = useState(controlledOpen ?? true)
```
tomas562, line 48: The open prop is only read during initial render. If a consumer changes it to false after mount, the component won't close. For standalone (non-provider) usage this is a hidden footgun. Maybe we could sync via useEffect?

src/components/common/Snackbar/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+      TransitionComponent={Slide}
+      transitionDuration={prefersReducedMotion ? { enter: 0, exit: 0 } : undefined}
+      sx={{
+        bottom: `${baseOffset + bottomOffset}px !important`,
+        transition: prefersReducedMotion ? 'none' : `bottom 150ms ease-in-out !important`,
+      }}
+    >
+      <Box
+        role='status'
+        aria-live={config.ariaLive}
```
tomas562: role='status' is semantically equivalent to aria-live='polite', so for error/action-needed states where ariaLive='assertive', the role overrides the intent — screen readers won't interrupt for error messages. Fix:
`role={config.ariaLive === 'assertive' ? 'alert' : 'status'}`

src/components/common/Snackbar/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+          transition: 'opacity 200ms ease-out',
+        }}
+      >
+        <Box
+          sx={{
+            flexShrink: 0,
+            display: 'flex',
+            alignItems: 'center',
+            justifyContent: 'center',
+            marginRight: theme.spacing(1),
```
tomas562: Double spacing between icon and text. The outer Box has gap: theme.spacing(1) (8px) AND the icon Box has marginRight: theme.spacing(1) (8px), giving 16px total between the icon and text. My suggestion: Set spacing in one place and remove marginRight from the icon wrapper Box.

src/components/common/Snackbar/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+    onAction?.()
+    if (dismissOnAction) handleClose()
+  }
+
+  const snackbarHeight = sizeConfig[size].height
+  const bottomOffset = index * (snackbarHeight + SNACKBAR_STACK_GAP)
+
+  const prefersReducedMotion =
+    typeof window !== 'undefined' &&
+    window.matchMedia('(prefers-reduced-motion: reduce)').matches
```
tomas562: You can use import { useMediaQuery } from '@mui/material'

`const prefersReducedMotion = useMediaQuery('(prefers-reduced-motion: reduce)')`

src/components/common/Snackbar/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+  const isMobile =
+    typeof window !== 'undefined' && window.innerWidth < theme.breakpoints.values.md
+  const baseOffset = isMobile ? SNACKBAR_BASE_OFFSET.mobile : SNACKBAR_BASE_OFFSET.desktop
```
tomas562: const isMobile = useMediaQuery(theme.breakpoints.down('md'))

src/components/common/Snackbar/config.ts
```diff
@@ -0,0 +1,83 @@
...
+export const MAX_SNACKBARS_DESKTOP = 3
+
+export const SNACKBAR_STACK_GAP = 8
+export const SNACKBAR_ICON_SIZE = { width: 18, height: 20 }
+export const SNACKBAR_BASE_OFFSET = { mobile: 16, desktop: 24 }
+
+export const stateConfig: Record<
+  SnackbarState,
+  {
+    icon: React.ComponentType<{ fillColor?: string; 'aria-hidden'?: boolean }>
```
tomas562: Instead of relying on an implicit global type we could import it:
` import type { ComponentType } from 'react'`

src/components/common/Snackbar/SnackbarProvider.tsx
```diff
@@ -0,0 +1,85 @@
...
+  duration?: number
+  persist?: boolean
+  dismissOnAction?: boolean
+}
+
+interface SnackbarContextType {
+  showSnackbar: (message: Omit<SnackbarMessage, 'id'>) => string
+  hideSnackbar: (id: string) => void
+  messages: SnackbarMessage[]
+}
```
tomas562, line 28: messages is on the public SnackbarContextType but consumers of useSnackbar() only need showSnackbar/hideSnackbar. Exposing the internal state leaks implementation detail. Could we remove it form exported interface?

src/components/common/Snackbar/SnackbarProvider.tsx
```diff
@@ -0,0 +1,85 @@
...
+          size={msg.size}
+          onClose={() => hideSnackbar(msg.id)}
+          onAction={msg.onAction}
+          dismissOnAction={msg.dismissOnAction}
+          actionLabel={msg.actionLabel}
+          duration={msg.duration}
+          persist={msg.persist}
+          index={index}
+          open
+        />
```
tomas562, line 64: RcSesSnackbar has a showClose prop (defaults to true), but SnackbarMessage interface doesn't include it - so consumers using the provider can never hide the close button.

src/components/common/Snackbar/index.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+    const { container } = renderSnackbar(
+      <RcSesSnackbar {...DEFAULT_PROPS} onClose={handleClose} />,
+    )
+    const snackbar = getStatusElement(container)
+    if (snackbar) {
+      fireEvent.keyDown(snackbar, { key: 'Escape', code: 'Escape' })
+    }
+
+    expect(handleClose).toHaveBeenCalled()
+  })
```
tomas562, line 73: The onKeyDown handler lives on MuiSnackbar, but the test fires it on role='status' (inner Box). It works via bubbling, but this is fragile. The test also misses coverage for: auto-hide after duration and persist behavior

src/components/common/Snackbar/SnackbarProvider.tsx
```diff
@@ -0,0 +1,85 @@
...
+  messages: SnackbarMessage[]
+}
+
+const SnackbarContext = createContext<SnackbarContextType | undefined>(undefined)
+
+export function RcSesSnackbarProvider({ children }: { children: React.ReactNode }) {
+  const [messages, setMessages] = useState<SnackbarMessage[]>([])
+
+  const showSnackbar = useCallback((message: Omit<SnackbarMessage, 'id'>) => {
+    const id = `snackbar-${Date.now()}-${Math.random()}`
```
tomas562: [crypto.randomUUID()](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) would be cleaner

saukaite: ✅ Updated

src/components/common/Snackbar/index.test.tsx
saukaite: ✅ Updated, the test now fires `keyDown` directly on MuiSnackbar element
✅ Added tests: 'auto-hides after duration' and 'does not auto-hide when persist is true'

src/components/common/Snackbar/SnackbarProvider.tsx
saukaite: ✅ Good catch, added missing `showClose` prop to the `SnackbarMessage`.

src/components/common/Snackbar/SnackbarProvider.tsx
saukaite: ✅ Good point, removed

src/components/common/Snackbar/config.ts
saukaite: ✅ Updated

src/components/common/Snackbar/index.tsx
saukaite: ✅ Updated

src/components/common/Snackbar/index.tsx
saukaite: ✅ Updated

src/components/common/Snackbar/index.tsx
saukaite: ✅ Fixed

src/components/common/Snackbar/index.tsx
saukaite: ✅ Updated

src/components/common/Snackbar/index.tsx
saukaite: ✅ Added

```
  useEffect(() => {
    if (controlledOpen !== undefined) {
      setInternalOpen(controlledOpen)
    }
  }, [controlledOpen])
```

tomas562 (approved):

PR #91, opened 2026-05-07
SAV-5565: update checkbox component

saukaite:
## 🔗 Task

https://jira.registrucentras.lt/jira/browse/SAV-5565

## 🧾 Summary

Added **Indeterminate (mixed)** state:
✅ If no child checkboxes are selected and the parent checkbox is checked, all child checkboxes become checked.
✅  If all child checkboxes are selected and the parent checkbox is unchecked, all child checkboxes become unchecked.
✅  If only some child checkboxes are selected, the parent checkbox should be **indeterminate (mixed state)**.
✅  If some child checkboxes are selected and the parent checkbox is clicked, all checkboxes become checked.

Missed one extra case, so added:
✅ If child values are unchecked one by one, parent checkbox becomes unchecked too

## 📸 Screenshots

Screenshot 2026-05-12 151155

Screenshot 2026-05-12 151207

## ✅ Checklist

* [x] Tests added/updated

## ⚠️ Risks

N/A

### DIFF
--- a/src/components/form/inputs/CheckboxFormControl.test.tsx
+++ b/src/components/form/inputs/CheckboxFormControl.test.tsx
@@ -1,0 +1,47 @@
+import { ThemeProvider } from '@mui/material/styles'
+import { render } from '@testing-library/react'
+import { useForm } from 'react-hook-form'
+
+import theme from '@/theme/light'
+
+import RcSesCheckboxFormControl from './CheckboxFormControl'
+
+type WrapperProps = Partial<React.ComponentProps<typeof RcSesCheckboxFormControl>> & {
+  defaultValue?: boolean
+}
+
+const TestWrapper = ({ defaultValue = false, ...props }: WrapperProps) => {
+  const { control } = useForm({ defaultValues: { checkbox: defaultValue } })
+  return (
+    <ThemeProvider theme={theme}>
+      <RcSesCheckboxFormControl
+        id='checkbox'
+        name='checkbox'
+        control={control}
+        {...props}
+      >
+        Test Checkbox
+      </RcSesCheckboxFormControl>
+    </ThemeProvider>
+  )
+}
+
+describe('RcSesCheckboxFormControl - Indeterminate State', () => {
+  test('should auto-check parent when all children are checked', () => {
+    const { container } = render(<TestWrapper childValues={[true, true, true]} />)
+    const checkbox = container.querySelector('input#checkbox') as HTMLInputElement
+    expect(checkbox.checked).toBe(true)
+  })
+
+  test('should auto-uncheck parent when all children are unchecked', () => {
+    const { container } = render(<TestWrapper childValues={[false, false, false]} />)
+    const checkbox = container.querySelector('input#checkbox') as HTMLInputElement
+    expect(checkbox.checked).toBe(false)
+  })
+
+  test('should render with mixed child values (indeterminate visual state)', () => {
+    const { container } = render(<TestWrapper childValues={[true, false, true]} />)
+    const checkbox = container.querySelector('input#checkbox')
+    expect(checkbox).toBeTruthy()
+  })
+})
--- a/src/components/form/inputs/CheckboxFormControl.test.tsx
+++ b/src/components/form/inputs/CheckboxFormControl.test.tsx
@@ -1,6 +1,7 @@
 import { ThemeProvider } from '@mui/material/styles'
 import { render } from '@testing-library/react'
 import { useForm } from 'react-hook-form'
+import { describe, expect, test, vi } from 'vitest'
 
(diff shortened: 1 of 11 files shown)


src/stories/components/form/CheckboxFormControl.stories.tsx
```diff
@@ -116,3 +117,101 @@ export const Main: Story = {
...
+              paddingLeft: '2rem',
+              marginTop: '0.75rem',
+              display: 'flex',
+              flexDirection: 'column',
+              gap: '0.5rem',
+            }}
+          >
+            <FormControlLabel
+              control={
+                <Checkbox
```
tomas562: Why don't we use RcSesCheckbox instead?

src/components/form/inputs/CheckboxFormControl.test.tsx
tomas562, line 1: it would be nice to have tests for:

- Clicking parent when indeterminate - all children become true (verifying onChildValuesChange is called with [true, true, true])

- Clicking checked parent - all children become false

src/components/form/inputs/CheckboxFormControl.tsx
```diff
@@ -59,26 +61,62 @@ const LoadingStateIcon = styled(SpinnerGapBoldIcon)(`
...
+
+    if (onChildValuesChange && childValues.length > 0) {
+      onChildValuesChange(childValues.map(() => newValue))
+    }
+  }
+
   return (
     <FormControlLabel
       control={
         <MuiCheckbox
```
tomas562: Why don't we use RcSesCheckbox instead?

tomas562 (commented):
Overall looks good. Missing states handling (probably this is out of scope for this task. Seems design is not fully confirmed. )

src/components/form/inputs/CheckboxFormControl.test.tsx
saukaite: Added two more tests ✅

src/stories/components/form/CheckboxFormControl.stories.tsx
saukaite: Used the new `SimpleCheckbox` component as discussed ✅

src/components/form/inputs/CheckboxFormControl.tsx
```diff
@@ -32,53 +31,72 @@ type Props = Pick<TControllerProps, ImmediateControllerProps> &
...
-     87.5% { transform: rotateZ(315deg); }
-    100.0% { transform: rotateZ(360deg); }
-  }
-
-  animation-duration: 1s;
-  animation-iteration-count: infinite;
-  animation-name: loadingStateIconRotation;
-  animation-timing-function: step-start;
-  box-shadow: none !important;
-`)
```
saukaite, line 59: This spinner is for the parent only, in the parent-child logic it looks weird as only parent checkbox is loading. Removed and added Skeleton from MUI.

src/components/form/inputs/CheckboxFormControl.tsx
```diff
@@ -32,53 +31,72 @@ type Props = Pick<TControllerProps, ImmediateControllerProps> &
...
+    onChange(newValue)
+
+    if (onChildValuesChange && childValues.length > 0) {
+      onChildValuesChange(childValues.map(() => newValue))
+    }
+  }
+
   return (
-    <FormControlLabel
+    <FormControlLabelWithLoading
```
saukaite: This is the same `FormControlLabel` component, just with added `loading` functionality, as the original component does not have such prop. Otherwise, only checkboxes are loading and text doesn't have any loading indicator.

src/components/utils/loading.tsx
```diff
@@ -0,0 +1,9 @@
+import { Skeleton } from '@mui/material'
+import React from 'react'
+
+const LOADING_SKELETON = <Skeleton variant='rounded' width={20} height={20} />
+
+const getLoadingIcon = (fallback: React.ReactNode, loading?: boolean) =>
+  loading ? LOADING_SKELETON : fallback
```
saukaite, line 7: Added this simple util to reduce repetition in a couple files where it's needed.

src/components/form/inputs/FormControlLabelWithLoading.tsx
```diff
@@ -0,0 +1,39 @@
...
+          />
+        ) : (
+          label
+        )
+      }
+    />
+  )
+}
+
+export default RcSesFormControlLabelWithLoading
```
tomas562: maybe just `RcSesFormControlLabel`?

saukaite: Simplified naming ✅

tomas562 (approved):

PR #99, opened 2026-05-21
SAV-5545: add Button loading state + LoadingSpinner component

saukaite:
## 🔗 Task

https://jira.registrucentras.lt/jira/browse/SAV-5704

## 🧾 Summary

**Spinner Placement:**

✅ Text only → spinner replaces text
✅ Text + icon → spinner replaces icon only, text stays
✅ Icon only → spinner replaces icon

**Accessibility:**

✅ aria-busy="true" during loading
✅ Screen reader announces "Kraunasi"/"Loading" (aria-label + aria-live)
✅ Button disabled during loading (click/Enter/Space ignored)
✅ Focus indicator preserved (MUI default)

**Animation:**

✅ 360° rotation per 1 second, linear, continuous 
✅ prefers-reduced-motion: reduce → 2 second animation

## 📸 Screenshots

Check in the Storybook: 
https://ses-react-storybook.registrucentras.lt/preview/SAV-5745-unify-button-loading-state/?path=/docs/components-actions-button--docs

Example:

image

## ✅ Checklist

* [x] Tests added/updated

## ⚠️ Risks

N/A


src/components/common/Button/index.tsx
```diff
@@ -14,21 +17,56 @@ const ICON_ONLY_SIZE_MAP = {
   large: '3rem',
 } as const
 
-type Props = ButtonProps & { to?: string }
```
saukaite, line 17: ❓ `to` prop was not used, not sure why needed?

src/components/common/Button/index.tsx
```diff
@@ -14,21 +17,56 @@ const ICON_ONLY_SIZE_MAP = {
...
+type Props = ButtonProps & {
+  loading?: boolean
+}
```
saukaite, line 24: In MUI v5 `MuiButton` doesn't have `loading` state, only from MUI v6.

src/components/common/Button/index.tsx
```diff
@@ -14,21 +17,56 @@ const ICON_ONLY_SIZE_MAP = {
...
+  const displayEndIcon = getIconDisplay(rest.endIcon)
+
+  let displayContent = children
+  if (showSpinnerInContent || showSpinnerAsIcon) displayContent = spinner
+
   return (
     <MuiButton
       {...defaultProps}
       {...rest}
+      disabled={loading || rest.disabled}
```
saukaite, line 65: I think it makes sense to disable button while loading (grey background is added) - it doesn't match Figma (no grey background there), but I think we should keep this way to indicate that button is not clickable - it's a common pattern.

tomas562: Agree. Let's keep it like this now. Will need to refactor a bit when  MUI +v6 will be introduced. Please add TODO as reminder and we need jira ticket in backlog as well

src/components/common/Button/index.tsx
tomas562: Will need to refactor when  MUI +v6 will be introduced. Please add TODO as reminder and we need jira ticket in backlog as well

src/components/common/Button/index.tsx
```diff
@@ -14,21 +17,56 @@ const ICON_ONLY_SIZE_MAP = {
...
       size={size}
-      variant={variant}
+      variant={currentVariant}
+      aria-busy={loading}
```
tomas562:
```suggestion
      aria-busy={loading || undefined}
```

src/components/common/Button/LoadingSpinner.tsx
```diff
@@ -0,0 +1,56 @@
...
+        },
+        '@media (prefers-reduced-motion: reduce)': {
+          animationDuration: '2s',
+        },
+      }}
+    />
+  )
+}
+
+export default LoadingSpinner
```
tomas562: component is not exported in src\library\index.ts

src/components/common/Button/index.tsx
```diff
@@ -14,21 +17,56 @@ const ICON_ONLY_SIZE_MAP = {
...
+      aria-live={loading ? 'polite' : undefined}
+      aria-label={loading ? t('components.Button.loading') : rest['aria-label']}
```
tomas562: aria-label destroys original accessible name when loading. Screen reader users lose context - "Loading" alone doesn't tell them which button is loading. Consider preserving the original label:

```suggestion
     aria-label={loading && rest['aria-label'] ? `${rest['aria-label']} – ${t('components.Button.loading')}` : rest['aria-label']}
```
Or omit the aria-label change entirely since aria-busy="true" already signals the loading state.

src/components/common/Button/index.tsx
```diff
@@ -14,21 +17,56 @@ const ICON_ONLY_SIZE_MAP = {
...
   return (
     <MuiButton
       {...defaultProps}
       {...rest}
+      disabled={loading || rest.disabled}
       size={size}
-      variant={variant}
+      variant={currentVariant}
+      aria-busy={loading}
+      aria-live={loading ? 'polite' : undefined}
```
tomas562: aria-live="polite" on a button tag is non-standard. aria-live is for live regions (container elements whose text content changes dynamically). Placing it on an interactive element has inconsistent support across screen readers.

src/components/LoadingSpinner/LoadingSpinner.tsx
```diff
@@ -0,0 +1,56 @@
...
+  color?: SpinnerColor
+}
+
+function LoadingSpinner({ color = 'inherit' }: LoadingSpinnerProps) {
+  return (
+    <CircularProgress
+      size='20px'
+      color={color}
+      sx={{
+        animation: 'spin 1s linear infinite',
```
tomas562, line 53: the custom animation: 'spin 1s linear' overrides MUI's circular-rotate 1.4s linear infinite on the wrapper, but the inner SVG <circle> still runs MUI's circular-dash 1.4s ease-in-out infinite. This means the rotation (1s) and stroke-dash (1.4s) run at mismatched speeds, which can look jittery. Either override both animations or use variant='determinate' value={75} with a single rotation to get a clean spinner.

src/components/common/Button/LoadingSpinner.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+interface LoadingSpinnerProps {
+  color?: SpinnerColor
+}
+
+function LoadingSpinner({ color = 'inherit' }: LoadingSpinnerProps) {
+  return (
+    <CircularProgress
+      size='20px'
```
tomas562: size hardcoded at 20px regardless of button size (If we think in button context
 const size: OverridableStringUnion<"small" | "medium" | "large", ButtonPropsSizeOverrides>)
maybe const SPINNER_SIZE_MAP = { small: '16px', medium: '20px', large: '24px' }. But I would not like idea to lok only on these. Maybe we could have custom option as well to enterany wanted walue..? This could be used for page loaders/other scenarios

src/components/common/Button/index.test.tsx
```diff
@@ -0,0 +1,161 @@
+import { fireEvent, render, screen } from '@testing-library/react'
+import { describe, expect, it, vi } from 'vitest'
+
+import Button from './index'
+
+describe('Button Component', () => {
+  describe('Loading State', () => {
```
tomas562, line 7: Good job! Small nit pick: no test for loading preventing click on its own

src/components/common/Button/LoadingSpinner.tsx
saukaite: Yes, I didn't export it as I thought we would use it only inside Button component, but can export 👌

src/components/common/Button/index.tsx
saukaite: 🧹

src/components/LoadingSpinner/LoadingSpinner.tsx
saukaite: Good catch 👍 updated

src/components/common/Button/LoadingSpinner.tsx
saukaite: Agree, will add 3 defined sizes and a string, so any value can be passed:

```
interface LoadingSpinnerProps {
  color?: SpinnerColor
  size?: 'small' | 'medium' | 'large' | string
}
```

src/components/common/Button/index.tsx
saukaite: Will keep updated aria-label 👌

src/components/common/Button/index.tsx
saukaite: Added reminder, pinged Algirdas regarding new Jira task

src/components/common/Button/index.test.tsx
saukaite: Added ✅

src/components/common/Button/LoadingSpinner.tsx
```diff
@@ -0,0 +1,67 @@
...
+} as const
+
+// Custom colors that aren't MUI standard map to inherit (uses parent color)
+const CUSTOM_COLOR_MAP: Record<string, SpinnerColor> = {
+  light: 'inherit',
+  ghost: 'inherit',
+  dark: 'inherit',
+}
+
+export function getSpinnerColor(color: any): SpinnerColor {
```
tomas562: any type? maybe we could use  ButtonOwnProps['color'] or something like that..
```suggestion
export function getSpinnerColor(color: ButtonOwnProps['color']): SpinnerColor {
```

src/components/common/Button/LoadingSpinner.tsx
```diff
@@ -0,0 +1,67 @@
...
+  if (VALID_SPINNER_COLORS.includes(color)) {
+    return color
+  }
+  return CUSTOM_COLOR_MAP[color] ?? 'inherit'
+}
+
+interface LoadingSpinnerProps {
```
tomas562: interface is not exported for consumers. we have src\library\types.ts for this

src/components/common/Button/LoadingSpinner.tsx
```diff
@@ -0,0 +1,67 @@
...
+  'primary',
+  'secondary',
+  'error',
+  'warning',
+  'info',
+  'success',
+  'inherit',
+] as const
+
+type SpinnerColor = (typeof VALID_SPINNER_COLORS)[number]
```
tomas562: type is not exported for consumers

src/components/common/Button/LoadingSpinner.tsx
tomas562: Component is ready and could live it's own life. No point to keep it private..
Also LoadingSpinner breaks RcSes* naming convention. Since we exporting it now it should have it's own directory and storybook..

saukaite: Makes sense, will add shortly 👍

saukaite: Renamed, changed dir, added to the storybook 👍 

image

src/components/common/Button/LoadingSpinner.tsx
saukaite: ✅

src/components/common/Button/LoadingSpinner.tsx
saukaite: ✅

src/components/common/Button/LoadingSpinner.tsx
saukaite: ✅

tomas562 (approved):
good job!

PR #100, opened 2026-05-22
SAV-5705: create Dialog component

saukaite:
## 🔗 Task

https://jira.registrucentras.lt/jira/browse/SAV-5705

## 🧾 Summary

This is generic, basic `Dialog` component which can have custom `title`, `children`, `actions`. There're 3 size options: `sm`, `md` (default), `lg`.

Exact styling as the task needs will be applied separately (as the new component) by using this Dialog component as the base.

## 📸 Screenshots

Storybook: https://ses-react-storybook.registrucentras.lt/preview/SAV-5705-create-reusable-dialog-component/?path=/docs/components-overlays-dialog--docs

image

## ✅ Checklist

* [x] Tests added/updated

## ⚠️ Risks

N/A as component is not used yet.

### DIFF
--- a/src/components/overlays/Dialog/index.test.tsx
+++ b/src/components/overlays/Dialog/index.test.tsx
@@ -43,6 +43,24 @@ describe('RcSesDialog', () => {
     expect(paper).toHaveStyle({ width: '600px' })
   })
 
+  it('should apply small size (480px)', () => {
+    render(
+      <RcSesDialog open size='sm' dialogTitle='Test Title' actions={<Button>Close</Button>} />,
+    )
+
+    const paper = document.querySelector('.MuiDialog-paper')
+    expect(paper).toHaveStyle({ width: '480px' })
+  })
+
+  it('should apply large size (800px)', () => {
+    render(
+      <RcSesDialog open size='lg' dialogTitle='Test Title' actions={<Button>Close</Button>} />,
+    )
+
+    const paper = document.querySelector('.MuiDialog-paper')
+    expect(paper).toHaveStyle({ width: '800px' })
+  })
+
   it('should render ReactNode as title', () => {
     render(
       <RcSesDialog
--- a/src/components/overlays/Dialog/index.test.tsx
+++ b/src/components/overlays/Dialog/index.test.tsx
@@ -1,0 +1,57 @@
+import { Button } from '@mui/material'
+import { render, screen } from '@testing-library/react'
+import { describe, expect, it } from 'vitest'
+
+import RcSesDialog from './index'
+
+describe('RcSesDialog', () => {
+  it('should render dialog title', () => {
+    render(<RcSesDialog open dialogTitle='Test Title' actions={<Button>Close</Button>} />)
+
+    expect(screen.getByText('Test Title')).toBeInTheDocument()
+  })
+
+  it('should render children when provided', () => {
+    render(
+      <RcSesDialog open dialogTitle='Test Title' actions={<Button>Close</Button>}>
+        Test Content
+      </RcSesDialog>,
+    )
+
+    expect(screen.getByText('Test Content')).toBeInTheDocument()
+  })
+
+  it('should not render DialogContent when children are not provided', () => {
+    render(<RcSesDialog open dialogTitle='Test Title' actions={<Button>Close</Button>} />)
+
+    const dialogContent = document.querySelector('.MuiDialogContent-root')
+    expect(dialogContent).not.toBeInTheDocument()
+  })
+
+  it('should render actions', () => {
+    render(
+      <RcSesDialog open dialogTitle='Test Title' actions={<Button>Submit</Button>} />,
+    )
+
+    expect(screen.getByText('Submit')).toBeInTheDocument()
+  })
+
+  it('should apply medium size by default', () => {
+    render(<RcSesDialog open dialogTitle='Test Title' actions={<Button>Close</Button>} />)
+
+    const paper = document.querySelector('.MuiDialog-paper')
+    expect(paper).toHaveStyle({ width: '600px' })
+  })
+
+  it('should render ReactNode as title', () => {
+    render(
+      <RcSesDialog
+        open
+        dialogTitle={<h2>Custom Title Element</h2>}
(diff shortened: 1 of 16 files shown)


src/components/overlays/Dialog/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+  dialogTitle,
+  children,
+  actions,
+  onClose,
+  size = 'md',
+  ...props
+}: RcSesDialogProps) {
+  const dialogWidth = SIZE_MAP[size]
+  const titleId = useId()
+  const contentId = useId()
```
saukaite, line 36: This allows to assign `id` to elements, so that screen readers can find it by `id` and read the content.

src/components/overlays/Dialog/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+  sm: '480px',
+  md: '600px',
+  lg: '800px',
+}
+
+export interface RcSesDialogProps extends Omit<MuiDialogProps, 'children' | 'title'> {
+  dialogTitle: string | ReactNode
+  children?: ReactNode
+  actions: ReactNode
+  size?: DialogSize
```
saukaite, line 23: I think `dialogTitle` and `actions` should always be required.

src/components/overlays/Dialog/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+}: RcSesDialogProps) {
+  const dialogWidth = SIZE_MAP[size]
+  const titleId = useId()
+  const contentId = useId()
+
+  return (
+    <MuiDialog
+      onClose={onClose}
+      {...props}
+      maxWidth={false}
```
saukaite, line 42: MUI's default `maxWidth` is 600px, need to set it to `false`, as we want to have `large` modal which is 800px.

src/components/overlays/Dialog/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+    >
+      <DialogTitle
+        id={titleId}
+        sx={{
+          display: 'flex',
+          justifyContent: 'space-between',
+          alignItems: 'flex-start',
+          fontSize: '24px',
+          lineHeight: '32px',
+          padding: '2rem 2rem 1rem !important',
```
saukaite, line 55: `!important` is ugly, but I need to override the styles and I don't want to change `MuiDialog.ts`, because as I understand `ses-ui` would inherit these styling overrides as they use MUI Dialog in custom components.

src/components/overlays/Dialog/index.tsx
tomas562: maybe not in generic component. we can go stricter path for components built on top of this component?

src/components/overlays/Dialog/index.tsx
tomas562: good point!

src/components/overlays/Dialog/index.tsx
tomas562: RcSesDialogProps extends MuiDialogProps, so consumers can pass PaperProps or maxWidth - but both are set after the spread and will silently win.
Maybe correct type would be 
```
export interface RcSesDialogProps extends Omit<MuiDialogProps, 'children' | 'title' | 'PaperProps' | 'maxWidth'>
```
didn't try it. Just idea looking at the code. plz investigate

src/components/overlays/Dialog/index.test.tsx
tomas562, line 1: Tests only cover the default md size
There are no tests for sm(480px) or lg(800px). The size-mapping logic is simple but worth a quick assertion to prevent regressions..

src/stories/components/overlays/Dialog.stories.tsx
```diff
@@ -0,0 +1,276 @@
...
+    layout: 'centered',
+  },
+  tags: ['autodocs'],
+}
+
+export default meta
+type Story = StoryObj<{
+  dialogTitle: string
+  size?: 'sm' | 'md' | 'lg'
+}>
```
tomas562: Story type loses full component props. Should be StoryObj<typeof RcSesDialog> so all component props appear in the Storybook controls panel.

src/components/overlays/Dialog/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  return (
+    <MuiDialog
+      onClose={onClose}
+      {...props}
+      maxWidth={false}
+      PaperProps={{ sx: { width: dialogWidth, borderRadius: '16px' } }}
+      aria-labelledby={titleId}
+      aria-describedby={children ? contentId : undefined}
+      aria-modal
```
tomas562: already set by MUI's Dialog internally

src/components/overlays/Dialog/index.tsx
tomas562:
```suggestion
const contentId = children ? useId() : undefined
```

src/components/overlays/Dialog/index.tsx
saukaite: But without title and actions component would not be functional at all? 🤔

saukaite: I'm thinking from the functionality perspective, which props are must for the component to be meaningful

src/components/overlays/Dialog/index.tsx
saukaite: Good point, makes sense to omit them so that these props can be controlled only by the component. If we see any need, we can later allow partial control of e.g. `PaperProps` .

TS error appears if omitted:

Image

src/components/overlays/Dialog/index.tsx
saukaite: Good catch, seems to be redundant then 🧹

src/components/overlays/Dialog/index.tsx
tomas562: I see some possible variations witch could be created from this component. but maybe you are right. It would be overengineering since there is no such need at the moment. We can easily make it optional in the future without any breaking changes

tomas562 (approved):
LGTM

PR #106, opened 2026-06-12
Update: Pencil icon weight option

TautvydasBliudzius:
## 🧾 Summary

Updated PencilSimpleLineIcon to support the weight prop and forward it to the underlying Phosphor icon component.

- Added weight prop support to PencilSimpleLineIcon.
- Changed the default icon weight from 'fill' to 'regular'.
- Icons can now be rendered with different Phosphor weights (e.g. regular, fill, bold, duotone) when needed.
- Preserved backward compatibility by using 'regular' as the fallback value when weight is not provided.

## 📸 Screenshots

image

## ✅ Checklist

* [ ] Tests added/updated
* [+] UI matches approved design (Figma)
* [+] Accessibility reviewed (keyboard navigation, labels, semantics)

## ⚠️ Risks

- Existing consumers that relied on the icon being rendered with the fill variant by default will now receive the regular variant unless they explicitly pass weight="fill".
- Minor visual differences may appear in places where the icon was previously expected to be filled.

### DIFF
--- a/src/assets/icons/PencilSimpleLineIcon.tsx
+++ b/src/assets/icons/PencilSimpleLineIcon.tsx
@@ -6,7 +6,7 @@ function PencilSimpleLineIcon({
   className,
   fillColor,
   size = 24,
-  weight,
+  weight = 'regular',
   ...rest
 }: IconProps) {
   return (
@@ -14,7 +14,7 @@ function PencilSimpleLineIcon({
       {...rest}
       className={className}
       size={size}
-      weight={weight || 'regular'}
+      weight={weight}
       color={fillColor}
     />
   )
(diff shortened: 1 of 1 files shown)


src/assets/icons/PencilSimpleLineIcon.tsx
```diff
@@ -2,13 +2,19 @@ import { PencilSimpleLineIcon as PencilSimpleLine } from '@phosphor-icons/react'
...
+}: IconProps) {
   return (
     <PencilSimpleLine
       {...rest}
       className={className}
       size={size}
-      weight='fill'
+      weight={weight || 'regular'}
       color={fillColor}
     />
```
tomas562, line 19: Let's keep same pattern like in other icons

```suggestion
function PencilSimpleLineIcon({
  className,
  fillColor,
  size = 24,
  weight = 'regular',
  ...rest
}: IconProps) {
  return (
    <PencilSimpleLine
      {...rest}
      className={className}
      size={size}
      weight={weight}
      color={fillColor}
    />
  )
}
```

tomas562 (approved):

PR #107, opened 2026-06-29
SAV-6142: add SegmentedControl component

saukaite:
## 🔗 Task

https://jira.registrucentras.lt/jira/browse/SAV-6142

## 🧾 Summary

Added a new `SegmentedControl` component which uses a new `SegmentButton` component. The button has two sizes: `regular` and `small`. `SegmentedControl` component has an option to disable bg color (grey) via `disableBg` prop.

## 📸 Screenshots

It's best to check directly in the Storybook (updated with the latest changes):

https://ses-react-storybook.registrucentras.lt/preview/SAV-6142-add-segmented-control-component/?path=/docs/components-display-segmentedcontrol--docs

image

image

## ✅ Checklist

* [x] Tests added/updated

## ⚠️ Risks

N/A, component is not used yet.

### DIFF
--- a/src/components/common/SegmentedControl/components/SegmentButton/index.test.tsx
+++ b/src/components/common/SegmentedControl/components/SegmentButton/index.test.tsx
@@ -1,0 +1,85 @@
+import { fireEvent, render, screen } from '@testing-library/react'
+import { describe, expect, test, vi } from 'vitest'
+
+import RcSesSegmentButton, { type RcSesSegmentOption } from './index'
+
+describe('RcSesSegmentButton', () => {
+  const defaultOption: RcSesSegmentOption = {
+    id: 'option-1',
+    label: 'Option 1',
+    disabled: false,
+  }
+
+  const defaultProps = {
+    option: defaultOption,
+    isSelected: false,
+    isFocused: false,
+    size: 'regular' as const,
+    ariaRole: 'radio' as const,
+    onClick: vi.fn(),
+    onKeyDown: vi.fn(),
+    onFocus: vi.fn(),
+  }
+
+  test('renders button wtesth option label', () => {
+    render(<RcSesSegmentButton {...defaultProps} />)
+    expect(screen.getByRole('radio', { name: 'Option 1' })).toBeInTheDocument()
+  })
+
+  test('calls onClick handler when clicked', () => {
+    const onClick = vi.fn()
+
+    render(<RcSesSegmentButton {...defaultProps} onClick={onClick} />)
+
+    fireEvent.click(screen.getByRole('radio'))
+    expect(onClick).toHaveBeenCalledWith('option-1')
+  })
+
+  test('calls onFocus handler when focused', () => {
+    const onFocus = vi.fn()
+
+    render(<RcSesSegmentButton {...defaultProps} onFocus={onFocus} />)
+
+    fireEvent.focus(screen.getByRole('radio'))
+    expect(onFocus).toHaveBeenCalled()
+  })
+
+  test('calls onKeyDown handler when key pressed', () => {
+    const onKeyDown = vi.fn()
+
+    render(<RcSesSegmentButton {...defaultProps} onKeyDown={onKeyDown} />)
+
+    const button = screen.getByRole('radio')
+    fireEvent.keyDown(button, { key: 'ArrowRight', code: 'ArrowRight' })
+    expect(onKeyDown).toHaveBeenCalled()
+  })
+
+  test('disables button when option.disabled is true', () => {
+    render(
+      <RcSesSegmentButton
+        {...defaultProps}
+        option={{ ...defaultOption, disabled: true }}
+      />,
+    )
+
+    expect(screen.getByRole('radio')).toBeDisabled()
+  })
+
+  test('sets aria-checked when selected', () => {
+    render(<RcSesSegmentButton {...defaultProps} isSelected />)
+
+    expect(screen.getByRole('radio')).toHaveAttribute('aria-checked', 'true')
+  })
+
+  test('sets tabIndex to 0 when focused', () => {
+    render(<RcSesSegmentButton {...defaultProps} isFocused />)
+
+    expect(screen.getByRole('radio')).toHaveAttribute('tabIndex', '0')
+  })
+
+  test('supports tab role', () => {
+    render(<RcSesSegmentButton {...defaultProps} ariaRole='tab' />)
+
+    expect(screen.getByRole('tab')).toBeInTheDocument()
+  })
+})
--- a/src/components/common/SegmentedControl/components/SegmentButton/index.test.tsx
+++ b/src/components/common/SegmentedControl/components/SegmentButton/index.test.tsx
@@ -21,7 +21,7 @@ describe('RcSesSegmentButton', () => {
     onFocus: vi.fn(),
   }
 
-  test('renders button wtesth option label', () => {
+  test('renders button with option label', () => {
     render(<RcSesSegmentButton {...defaultProps} />)
     expect(screen.getByRole('radio', { name: 'Option 1' })).toBeInTheDocument()
   })
--- a/src/components/common/SegmentedControl/components/SegmentButton/index.tsx
+++ b/src/components/common/SegmentedControl/components/SegmentButton/index.tsx
@@ -5,14 +5,14 @@ import { grey, primary } from '@/theme/palette'
 
 const SIZE_STYLES: Record<'small' | 'regular', React.CSSProperties> = {
   small: {
-    width: '160px',
+    width: 'auto',
     minHeight: '36px',
     padding: '8px 12px',
     fontSize: '14px',
     lineHeight: '20px',
   },
   regular: {
-    width: '180px',
+    width: 'auto',
     minHeight: '44px',
     padding: '10px 16px',
     fontSize: '15px',
--- a/src/components/common/SegmentedControl/components/SegmentButton/index.tsx
+++ b/src/components/common/SegmentedControl/components/SegmentButton/index.tsx
@@ -88,8 +88,8 @@ const RcSesSegmentButton = React.forwardRef<HTMLButtonElement, RcSesSegmentButto
         onFocus={onFocus}
         disabled={isDisabled}
         role={ariaRole}
-        aria-checked={isSelected}
-        aria-selected={isSelected}
+        aria-checked={ariaRole === 'radio' ? isSelected : undefined}
+        aria-selected={ariaRole === 'tab' ? isSelected : undefined}
         tabIndex={isFocused && !isDisabled ? 0 : -1}
         sx={{
           ...buttonStyles,
--- a/src/components/common/SegmentedControl/components/SegmentButton/index.tsx
+++ b/src/components/common/SegmentedControl/components/SegmentButton/index.tsx
@@ -26,6 +26,7 @@ export interface RcSesSegmentOption {
   id: string
   label: string
   disabled?: boolean
+  panelId?: string
 }
 
 export interface RcSesSegmentButtonProps {
@@ -92,6 +93,7 @@ const RcSesSegmentButton = React.forwardRef<HTMLButtonElement, RcSesSegmentButto
         role={ariaRole}
         aria-checked={ariaRole === 'radio' ? isSelected : undefined}
         aria-selected={ariaRole === 'tab' ? isSelected : undefined}
+        aria-controls={option.panelId}
         tabIndex={isFocused && !isDisabled ? 0 : -1}
         sx={{
           ...buttonStyles,
--- a/src/components/common/SegmentedControl/components/SegmentButton/index.tsx
+++ b/src/components/common/SegmentedControl/components/SegmentButton/index.tsx
@@ -1,0 +1,107 @@
+import { Button as MuiButton } from '@mui/material'
+import React, { useCallback, useMemo } from 'react'
+
+import { grey, primary } from '@/theme/palette'
+
+const SIZE_STYLES: Record<'small' | 'regular', React.CSSProperties> = {
+  small: {
+    width: '160px',
+    minHeight: '36px',
+    padding: '8px 12px',
+    fontSize: '14px',
+    lineHeight: '20px',
+  },
+  regular: {
+    width: '180px',
+    minHeight: '44px',
+    padding: '10px 16px',
+    fontSize: '15px',
+    lineHeight: '20px',
+  },
+}
+
+export interface RcSesSegmentOption {
+  id: string
+  label: string
+  disabled?: boolean
+}
+
+export interface RcSesSegmentButtonProps {
+  option: RcSesSegmentOption
+  isSelected: boolean
+  isFocused: boolean
+  size: 'small' | 'regular'
+  ariaRole: 'tab' | 'radio'
+  onClick: (optionId: string) => void
+  onKeyDown: (e: React.KeyboardEvent<HTMLButtonElement>) => void
+  onFocus: () => void
+}
+
+const RcSesSegmentButton = React.forwardRef<HTMLButtonElement, RcSesSegmentButtonProps>(
+  (
+    { option, isSelected, isFocused, size, ariaRole, onClick, onKeyDown, onFocus },
+    ref,
+  ) => {
+    const isDisabled = option.disabled
+    const sizeStyles = SIZE_STYLES[size]
+
+    const textColor = useMemo(() => {
+      if (isDisabled) return `${grey['500']} !important`
+      if (isSelected) return `${grey['900']} !important`
+      return `${primary['700']} !important`
+    }, [isDisabled, isSelected])
+
+    const buttonStyles = useMemo(
+      () =>
+        ({
+          flex: '0 1 auto',
+          minWidth: '0',
+          height: 'auto',
+          wordBreak: 'break-word',
+          fontFamily: '"Public Sans", sans-serif !important',
+          border: `2px solid ${isFocused && !isDisabled ? primary['500'] : 'transparent'}`,
+          backgroundColor: isSelected
+            ? `${primary['50']} !important`
+            : 'transparent !important',
+          color: textColor,
+          cursor: isDisabled ? 'not-allowed' : 'pointer',
+          fontWeight: 500,
+          borderRadius: '8px',
+          transition: 'all 150ms ease-in-out',
+          whiteSpace: 'normal',
+          outline: 'none',
+          ...sizeStyles,
+        }) as React.CSSProperties,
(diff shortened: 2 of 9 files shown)


src/components/common/SegmentedControl/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+    const containerRef = useRef<HTMLDivElement>(null)
+
+    useImperativeHandle(ref, () => containerRef.current as HTMLDivElement)
+
+    const getEnabledIndices = useCallback(
+      () => options.map((opt, i) => (!opt.disabled ? i : -1)).filter((i) => i !== -1),
+      [options],
+    )
+
+    const handleKeyDown = useCallback(
```
tomas562, line 65: Arrow/Home/End handlers update `focusIndex` state only — button refs are never attached and `.focus()` is never called, so DOM focus stays on the originally focused button. The focus border and `tabIndex` move visually, but the next keydown still fires on the old button: after ArrowRight + Enter the *previous* option gets selected, and screen readers announce nothing during navigation. Add a `useRef` array of button elements, attach them in the `options.map`, and call `buttonRefs.current[newIndex]?.focus()` in each navigation branch.

src/components/common/SegmentedControl/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+      options,
+      value,
+      onChange,
+      size = 'regular',
+      role = 'radiogroup',
+      disableBg = false,
+    },
+    ref,
+  ) => {
+    const [focusIndex, setFocusIndex] = useState(0)
```
tomas562: `focusIndex` starts at 0 regardless of `value`: tabbing into the group focuses the first option instead of the selected one, and when the first option is disabled every button gets `tabIndex={-1}`, making the whole control unreachable by keyboard. Initialize from the index of `value`, falling back to the first enabled option.

src/components/common/SegmentedControl/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+            e.preventDefault()
+            setFocusIndex(enabledIndices[0])
+            break
+
+          case 'End':
+            e.preventDefault()
+            setFocusIndex(enabledIndices[enabledIndices.length - 1])
+            break
+
+          case 'Enter':
```
tomas562, line 111: SAV-6142 AC #2 specifies arrows move the *selection* („perkelia pasirinkimą“), which is also the WAI-ARIA APG pattern for `radiogroup`. The implementation only moves focus and requires Enter/Space to select. For `role='radiogroup'`, call `onChange` in the arrow branches (selection follows focus), or align the deviation with the DS team.

src/components/common/SegmentedControl/components/SegmentButton/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+    return (
+      <MuiButton
+        ref={ref}
+        type='button'
+        onClick={handleClick}
+        onKeyDown={onKeyDown}
+        onFocus={onFocus}
+        disabled={isDisabled}
+        role={ariaRole}
+        aria-checked={isSelected}
```
tomas562: `aria-checked` and `aria-selected` are both always rendered; `aria-selected` is invalid on `role='radio'` and `aria-checked` is invalid on `role='tab'`. Render only the attribute matching `ariaRole`.

src/components/common/SegmentedControl/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+        borderRadius: '8px',
+      }),
+      [disableBg],
+    )
+
+    return (
+      <div
+        ref={containerRef}
+        role={role}
+        aria-label='Segmented control options'
```
tomas562: `aria-label='Segmented control options'` is hardcoded English in a Lithuanian-facing portal. Expose it as a prop so consumers pass a localized label.

src/components/common/SegmentedControl/components/SegmentButton/index.tsx
```diff
@@ -0,0 +1,107 @@
+import { Button as MuiButton } from '@mui/material'
+import React, { useCallback, useMemo } from 'react'
+
+import { grey, primary } from '@/theme/palette'
+
+const SIZE_STYLES: Record<'small' | 'regular', React.CSSProperties> = {
```
tomas562, line 6: The ticket notes the design changed from fixed-width segments to hug-content + wrap („hug segmentai + layoutWrap“), but `SIZE_STYLES` pins `width: 160px/180px`, so short labels pad to full width instead of hugging. Check against the current Figma component (node 47125-36) and switch to auto width if the spec is hug.

src/components/common/SegmentedControl/index.test.tsx
```diff
@@ -0,0 +1,157 @@
...
+    expect(screen.getByRole('radio', { name: 'Option 3' })).toBeInTheDocument()
+  })
+
+  test('has correct container role (radiogroup by default)', () => {
+    render(<RcSesSegmentedControl {...defaultProps} />)
+
+    expect(screen.getByRole('radiogroup')).toBeInTheDocument()
+  })
+
+  test('marks selected option wtesth aria-checked=true', () => {
```
tomas562: Find/replace artifact: "wtesth" appears in several test names and in `optionsWtesthDisabled` in both this file and the SegmentButton test file — rename back to "with".

src/components/common/SegmentedControl/index.test.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+  test('calls onChange when option clicked', () => {
+    const onChange = vi.fn()
+    render(<RcSesSegmentedControl {...defaultProps} onChange={onChange} />)
+
+    fireEvent.click(screen.getByRole('radio', { name: 'Option 2' }))
+    expect(onChange).toHaveBeenCalledWith('option-2')
+  })
+
+  test('navigates wtesth ArrowRight key', () => {
```
tomas562: The ArrowRight, Home/End, and skips-disabled tests only assert `toBeInTheDocument()` on an already-rendered element, so they pass regardless of navigation behavior — which is how the focus issue above slipped through. Once real focus management is in, assert the target button with `toHaveFocus()` and check `tabIndex` movement.

src/stories/components/display/SegmentedControl.stories.tsx
```diff
@@ -0,0 +1,388 @@
...
+    disableBg: {
+      control: 'boolean',
+      description: 'Disable container background',
+    },
+  },
+} satisfies Meta<typeof RcSesSegmentedControl>
+
+export default meta
+
+const DefaultComponent = ({
```
tomas562, line 45: Story renderers type props as `any` and each story logs to console behind an eslint-disable. One parameterized template component would remove the duplication; drop the console noise before merge.

tomas562 (changes requested):
Solid component structure and clean API; export wiring and styling conventions match the library. Requesting changes for: (1) keyboard navigation never moves DOM focus, so arrow + Enter selects the wrong option; (2) initial focus state ignores `value` and can make the control keyboard-unreachable; (3) arrow keys don't change selection as SAV-6142 AC #2 requires. The keyboard tests should assert focus so the fix is verified. Remaining comments are smaller a11y/spec/cleanup items.

src/components/common/SegmentedControl/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+      value,
+      onChange,
+      size = 'regular',
+      role = 'radiogroup',
+      disableBg = false,
+    },
+    ref,
+  ) => {
+    const [focusIndex, setFocusIndex] = useState(0)
+    const containerRef = useRef<HTMLDivElement>(null)
```
tomas562: Initializes focus from `value` (falling back to the first enabled option) and adds the button ref array needed to move real DOM focus. Apply together with the keydown-handler and ref-attachment suggestions below.

```suggestion
    const [focusIndex, setFocusIndex] = useState(() => {
      const selectedIndex = options.findIndex((opt) => opt.id === value && !opt.disabled)
      if (selectedIndex !== -1) return selectedIndex
      return Math.max(0, options.findIndex((opt) => !opt.disabled))
    })
    const containerRef = useRef<HTMLDivElement>(null)
    const buttonRefs = useRef<(HTMLButtonElement | null)[]>([])
```

tomas562 (commented):
Follow-up to my review: committable suggestions for the keyboard/ARIA findings. The three focus-management suggestions (initial state + refs, keydown handler, ref attachment) depend on each other - batch them into one commit. Same for the three aria-label ones.

src/components/common/SegmentedControl/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+            e.preventDefault()
+            onChange(options[index].id)
+            break
+
+          default:
+            break
+        }
+      },
+      [getEnabledIndices, onChange, options],
+    )
```
tomas562, line 122: Moves real DOM focus on navigation and makes arrows/Home/End move the selection per SAV-6142 AC #2 (selection follows focus, the APG radiogroup pattern). Depends on the `buttonRefs` suggestion above.

```suggestion
    const focusSegment = useCallback((index: number) => {
      setFocusIndex(index)
      buttonRefs.current[index]?.focus()
    }, [])

    const handleKeyDown = useCallback(
      (e: React.KeyboardEvent<HTMLButtonElement>, index: number) => {
        const enabledIndices = getEnabledIndices()
        const currentPos = enabledIndices.indexOf(index)

        switch (e.key) {
          case 'ArrowLeft':
          case 'ArrowUp': {
            e.preventDefault()
            const prevIndex =
              currentPos > 0
                ? enabledIndices[currentPos - 1]
                : enabledIndices[enabledIndices.length - 1]
            focusSegment(prevIndex)
            onChange(options[prevIndex].id)
            break
          }

          case 'ArrowRight':
          case 'ArrowDown': {
            e.preventDefault()
            const nextIndex =
              currentPos < enabledIndices.length - 1
                ? enabledIndices[currentPos + 1]
                : enabledIndices[0]
            focusSegment(nextIndex)
            onChange(options[nextIndex].id)
            break
          }

          case 'Home': {
            e.preventDefault()
            const firstIndex = enabledIndices[0]
            focusSegment(firstIndex)
            onChange(options[firstIndex].id)
            break
          }

          case 'End': {
            e.preventDefault()
            const lastIndex = enabledIndices[enabledIndices.length - 1]
            focusSegment(lastIndex)
            onChange(options[lastIndex].id)
            break
          }

          case 'Enter':
          case ' ':
            e.preventDefault()
            onChange(options[index].id)
            break

          default:
            break
        }
      },
      [focusSegment, getEnabledIndices, onChange, options],
    )
```

src/components/common/SegmentedControl/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+    return (
+      <div
+        ref={containerRef}
+        role={role}
+        aria-label='Segmented control options'
+        style={containerStyles}
+      >
+        {options.map((option, index) => (
+          <RcSesSegmentButton
+            key={option.id}
```
tomas562, line 169: Attaches the button refs so `focusSegment` can move DOM focus. Depends on the `buttonRefs` suggestion above.

```suggestion
          <RcSesSegmentButton
            key={option.id}
            ref={(el) => {
              buttonRefs.current[index] = el
            }}
```

src/components/common/SegmentedControl/components/SegmentButton/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+      <MuiButton
+        ref={ref}
+        type='button'
+        onClick={handleClick}
+        onKeyDown={onKeyDown}
+        onFocus={onFocus}
+        disabled={isDisabled}
+        role={ariaRole}
+        aria-checked={isSelected}
+        aria-selected={isSelected}
```
tomas562: Renders only the ARIA state attribute valid for the current role.

```suggestion
        aria-checked={ariaRole === 'radio' ? isSelected : undefined}
        aria-selected={ariaRole === 'tab' ? isSelected : undefined}
```

src/components/common/SegmentedControl/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+export type { RcSesSegmentOption }
+
+export interface RcSesSegmentedControlProps {
+  options: RcSesSegmentOption[]
+  value: string
+  onChange: (value: string) => void
+  size?: 'small' | 'regular'
+  role?: 'radiogroup' | 'tablist'
+  disableBg?: boolean
```
tomas562, line 16: Part 1/3 of the localized aria-label change: add the prop to the interface.

```suggestion
  disableBg?: boolean
  ariaLabel?: string
```

src/components/common/SegmentedControl/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+  RcSesSegmentedControlProps
+>(
+  (
+    {
+      options,
+      value,
+      onChange,
+      size = 'regular',
+      role = 'radiogroup',
+      disableBg = false,
```
tomas562, line 31: Part 2/3: destructure the new prop.

```suggestion
      disableBg = false,
      ariaLabel,
```

src/components/common/SegmentedControl/index.tsx
tomas562: Part 3/3: use the consumer-provided label instead of hardcoded English.

```suggestion
        aria-label={ariaLabel}
```

src/components/common/SegmentedControl/index.tsx
saukaite: fixed

src/components/common/SegmentedControl/index.tsx
saukaite: fixed

src/components/common/SegmentedControl/index.tsx
saukaite: fixed

src/components/common/SegmentedControl/components/SegmentButton/index.tsx
saukaite: fixed

src/components/common/SegmentedControl/index.tsx
saukaite: fixed

src/components/common/SegmentedControl/components/SegmentButton/index.tsx
saukaite: fixed in the `SAV-6142: fix stories, tests, etc.` commit

src/components/common/SegmentedControl/index.test.tsx
saukaite: fixed in the `SAV-6142: fix stories, tests, etc.` commit

src/stories/components/display/SegmentedControl.stories.tsx
saukaite: fixed in the `SAV-6142: fix stories, tests, etc.` commit

src/components/common/SegmentedControl/index.test.tsx
saukaite: fixed in the `SAV-6142: fix stories, tests, etc.` commit

src/components/common/SegmentedControl/components/SegmentButton/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+          backgroundColor: isSelected
+            ? `${primary['50']} !important`
+            : 'transparent !important',
+          color: textColor,
+          cursor: isDisabled ? 'not-allowed' : 'pointer',
+          fontWeight: 500,
+          borderRadius: '8px',
+          transition: 'all 150ms ease-in-out',
+          whiteSpace: 'normal',
+          outline: 'none',
```
tomas562: Removing the native outline leaves keyboard users relying solely on the `border` above (line 64), which is derived from `isFocused`/`focusIndex` state rather than real DOM focus — so it renders without focus and lingers after blur. Dropping `outline: 'none'` restores a focus-tracking indicator. (If the border is intended as a *selected* marker, key it off `isSelected` instead.)

```suggestion
          whiteSpace: 'normal',
```

src/components/common/SegmentedControl/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+      if (selectedIndex !== -1) return selectedIndex
+      return Math.max(
+        0,
+        options.findIndex((opt) => !opt.disabled),
+      )
+    })
+    const containerRef = useRef<HTMLDivElement>(null)
+    const buttonRefs = useRef<(HTMLButtonElement | null)[]>([])
+
+    useImperativeHandle(ref, () => containerRef.current as HTMLDivElement)
```
tomas562, line 48: `focusIndex` is only computed in the `useState` initializer, so it never re-syncs when the parent changes `value`/`options` programmatically. Selection follows `value`, but the roving `tabIndex={0}` stays on the stale index, so keyboard entry can land on the wrong segment. Add an effect to resync (using `React.useEffect` to avoid touching the import):

```suggestion
    useImperativeHandle(ref, () => containerRef.current as HTMLDivElement)

    React.useEffect(() => {
      const selectedIndex = options.findIndex((opt) => opt.id === value && !opt.disabled)
      if (selectedIndex !== -1) setFocusIndex(selectedIndex)
    }, [value, options])
```

src/components/common/SegmentedControl/components/SegmentButton/index.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+    isSelected: false,
+    isFocused: false,
+    size: 'regular' as const,
+    ariaRole: 'radio' as const,
+    onClick: vi.fn(),
+    onKeyDown: vi.fn(),
+    onFocus: vi.fn(),
+  }
+
+  test('renders button wtesth option label', () => {
```
tomas562: Typo in the test description (find/replace artifact).

```suggestion
  test('renders button with option label', () => {
```

tomas562 (commented):
Solid component with good Storybook/test coverage; low risk while it's unused. Two substantive items inline (roving-tabindex resync, keyboard focus visibility) plus a test-name typo. Non-blocking follow-ups for later: the `!important`-heavy styling diverges from the rest of `components/common/` and blocks consumer overrides (`styled()`/theme override would be cleaner), `role='tablist'` has no `aria-controls`/panel wiring, and a default accessible name (`ariaLabel`) is missing.

saukaite:
> : the !important-heavy styling diverges from the rest of components/common/ and blocks consumer overrides (styled()/theme override would be cleaner)

Agree, but since button component is used, styling in theme would affect regular button component which is not wanted. 

> role='tablist' has no aria-controls/panel wiring, and a default accessible name (ariaLabel) is missing.

✅ fixed in the last two commits (before 'Apply suggestions from code review')

tomas562 (approved):

PR #108, opened 2026-07-01
SAV-6111: fix SelectableCardListItem focus issue

saukaite:
## 🔗 Task

https://jira.registrucentras.lt/jira/browse/SAV-6111

## 🧾 Summary

Added `borderColor` for the `&:focus-visible` and ensured focus work correctly with the TAB.

## 📸 Screenshots

Updates can be checked here (updated!):

https://ses-react-storybook.registrucentras.lt/preview/SAV-6111-add-border-color-on-focus/

Black border appears when item is on focus. Pressing ENTER selects the radio button.

image

## ✅ Checklist

* [x] Tests added/updated

## ⚠️ Risks

N/A

### DIFF
--- a/src/components/common/SelectableCardList/components/SelectableCardListItem.test.tsx
+++ b/src/components/common/SelectableCardList/components/SelectableCardListItem.test.tsx
@@ -1,5 +1,5 @@
 import { ThemeProvider } from '@mui/material/styles'
-import { render, screen } from '@testing-library/react'
+import { fireEvent, render, screen } from '@testing-library/react'
 import { type ReactElement } from 'react'
 import { describe, expect, test, vi } from 'vitest'
 
@@ -93,5 +93,80 @@ describe('SelectableCardListItem', () => {
 
     expect(screen.getByText('List entry A')).toBeInTheDocument()
     expect(screen.getByText('List entry B')).toBeInTheDocument()
+  })
+
+  test('calls onSelect once when Enter key is pressed on radio', () => {
+    const onSelect = vi.fn()
+    renderItem(
+      <SelectableCardListItem
+        title='Test'
+        selected={false}
+        onSelect={onSelect}
+        listItems={[]}
+      />,
+    )
+
+    const radio = screen.getByRole('radio')
+    radio.focus()
+    fireEvent.keyDown(radio, { key: 'Enter', code: 'Enter' })
+
+    expect(onSelect).toHaveBeenCalledTimes(1)
+  })
+
+  test('calls onSelect once when radio is clicked (no double-call)', () => {
+    const onSelect = vi.fn()
+    renderItem(
+      <SelectableCardListItem
+        title='Test'
+        selected={false}
+        onSelect={onSelect}
+        listItems={[]}
+      />,
+    )
+
+    const radio = screen.getByRole('radio')
+    fireEvent.click(radio)
+
+    expect(onSelect).toHaveBeenCalledTimes(1)
+  })
+
+  test('calls onSelect once when card is clicked outside radio', () => {
+    const onSelect = vi.fn()
+    renderItem(
+      <SelectableCardListItem
+        title='Test title'
+        selected={false}
+        onSelect={onSelect}
+        listItems={[]}
+      />,
+    )
+
+    // click on the title area (outside the radio)
+    const titleElement = screen.getByRole('heading', { name: 'Test title' })
+    fireEvent.click(titleElement)
+
+    expect(onSelect).toHaveBeenCalledTimes(1)
+  })
+
+  test('loading card does not call onSelect when clicked', () => {
+    const onSelect = vi.fn()
+    const { container } = renderItem(
+      <SelectableCardListItem
+        title='Loading'
+        selected={false}
+        onSelect={onSelect}
+        isLoading
+        listItems={[]}
+      />,
+    )
+
+    // get the card container and try to click it
+    const cardContent = container.querySelector('[class*="MuiCardContent"]')
+    if (cardContent) {
+      fireEvent.click(cardContent)
+    }
+
+    // should not call onSelect due to pointerEvents: 'none'
+    expect(onSelect).not.toHaveBeenCalled()
   })
 })
--- a/src/components/common/SelectableCardList/components/SelectableCardListItem.test.tsx
+++ b/src/components/common/SelectableCardList/components/SelectableCardListItem.test.tsx
@@ -148,7 +148,7 @@ describe('SelectableCardListItem', () => {
     expect(onSelect).toHaveBeenCalledTimes(1)
   })
 
-  test('loading card does not call onSelect when clicked', () => {
+  test('loading card is not selectable', () => {
     const onSelect = vi.fn()
     const { container } = renderItem(
       <SelectableCardListItem
@@ -160,13 +160,11 @@ describe('SelectableCardListItem', () => {
       />,
     )
 
-    // get the card container and try to click it
-    const cardContent = container.querySelector('[class*="MuiCardContent"]')
-    if (cardContent) {
-      fireEvent.click(cardContent)
-    }
+    // no radio is rendered while loading, so there is no keyboard tab stop
+    expect(screen.queryByRole('radio')).toBeNull()
 
-    // should not call onSelect due to pointerEvents: 'none'
-    expect(onSelect).not.toHaveBeenCalled()
+    // mouse clicks are blocked by pointer-events on the card root
+    const card = container.querySelector('.MuiCard-root')
+    expect(card).toHaveStyle('pointer-events: none')
   })
 })
--- a/src/components/common/SelectableCardList/components/SelectableCardListItem.tsx
+++ b/src/components/common/SelectableCardList/components/SelectableCardListItem.tsx
@@ -33,11 +33,21 @@ const SelectableCardListItem = ({
       <Radio
         aria-label={`Select ${title}`}
         checked={selected}
+        onChange={onSelect}
+        onKeyDown={(e) => {
+          if (e.key === 'Enter') {
+            e.preventDefault()
+            onSelect()
+          }
+        }}
         sx={{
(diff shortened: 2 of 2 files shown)


src/components/common/SelectableCardList/components/SelectableCardListItem.tsx
```diff
@@ -70,16 +80,22 @@ const SelectableCardListItem = ({
   )
   return (
     <RcSesCard
+      component='button'
```
tomas562: A `<button>` may not contain interactive descendants or headings, and this one wraps the Radio `<input>`, an `<h3>`, and the list content — invalid HTML, and screen readers flatten it into one confusing control. It also defaults to `type="submit"`, so inside a form a card click submits the form. Drop `component='button'` and `tabIndex` and keep the card a `div` — the Radio is natively tabbable and `:focus-within` already provides the indicator; if the button stays for some reason, add `type='button'`.

src/components/common/SelectableCardList/components/SelectableCardListItem.tsx
```diff
@@ -70,16 +80,22 @@ const SelectableCardListItem = ({
...
       onClick={onSelect}
+      tabIndex={isLoading ? 0 : -1}
```
tomas562: This condition is inverted: the loading card (no radio, `pointerEvents: 'none'`) becomes a tab stop, while the interactive card is removed from the tab order. `pointer-events: none` doesn't block keyboard activation, so Tab + Enter on a loading card fires `onSelect()` mid-load. Tab currently works on interactive cards only because the inner Radio is the tab stop. Set this to `-1` unconditionally (or drop it together with `component='button'` — see comment above).

src/components/common/SelectableCardList/components/SelectableCardListItem.tsx
```diff
@@ -70,16 +80,22 @@ const SelectableCardListItem = ({
...
       sx={{
         cursor: isLoading ? 'default' : 'pointer',
         pointerEvents: isLoading ? 'none' : 'auto',
         border: '0.125rem solid',
         borderRadius: '0.75rem',
         padding: { xs: '1rem' },
         borderColor: selected ? palette.primary.main : palette.grey[300],
-        transition: 'all 0.2s ease',
+        transition: 'borderColor 0.2s ease, outline 0.2s ease',
```
tomas562: `transition` is emitted as raw CSS, so the property name must be kebab-case — `borderColor` never matches and the border stopped animating (the previous `all 0.2s ease` worked). The `outline` entry is a no-op since outline is always `none`. Change to `transition: 'border-color 0.2s ease'`.

src/components/common/SelectableCardList/components/SelectableCardListItem.tsx
```diff
@@ -33,11 +33,21 @@ const SelectableCardListItem = ({
       <Radio
         aria-label={`Select ${title}`}
         checked={selected}
+        onChange={onSelect}
```
tomas562: Clicking (or pressing Space on) the Radio now calls `onSelect` twice: once via `onChange`, and again when the click bubbles to the card's `onClick`. Safe only if every consumer's handler is idempotent. Stop propagation in the Radio handlers, or make the Radio the single interactive element.

src/components/common/SelectableCardList/components/SelectableCardListItem.tsx
```diff
@@ -70,16 +80,22 @@ const SelectableCardListItem = ({
...
         cursor: isLoading ? 'default' : 'pointer',
         pointerEvents: isLoading ? 'none' : 'auto',
         border: '0.125rem solid',
         borderRadius: '0.75rem',
         padding: { xs: '1rem' },
         borderColor: selected ? palette.primary.main : palette.grey[300],
-        transition: 'all 0.2s ease',
+        transition: 'borderColor 0.2s ease, outline 0.2s ease',
         gap: { xs: '0.5rem', md: '0.25rem' },
+        '&:focus-visible, &:focus-within': {
```
tomas562: `:focus-within` also matches after a mouse click, so a just-clicked selected card shows the grey[900] border instead of `primary.main` until focus moves away. For a keyboard-only indicator use `&:has(input:focus-visible)` (mouse-clicked radios don't match `:focus-visible`). The `&:focus-visible` half is dead code once the card itself isn't focusable.

tomas562 (commented):
Goal is right and the `:focus-within` border works, but the `component='button'` + `tabIndex` combination introduces regressions. The Radio input was already Tab-focusable before this change — the missing pieces were only the visible focus style and the Enter handler, so most of this can be simplified. Details inline.

Tests: none of the new behavior is covered — add cases for Enter selecting once, click selecting once, and loading items not being selectable.

src/components/common/SelectableCardList/components/SelectableCardListItem.tsx
saukaite: `component='button'` + `tabIndex` - both removed, card is a semantic [div]

src/components/common/SelectableCardList/components/SelectableCardListItem.tsx
saukaite: `tabIndex={isLoading ? 0 : -1})` - removed entirely

src/components/common/SelectableCardList/components/SelectableCardListItem.tsx
saukaite: Makes sense, fixed to kebab-case

src/components/common/SelectableCardList/components/SelectableCardListItem.tsx
saukaite: Added `stopPropagation()`

src/components/common/SelectableCardList/components/SelectableCardListItem.tsx
saukaite: Changed to `'&:has(input:focus-visible)'`

saukaite:
> Goal is right and the `:focus-within` border works, but the `component='button'` + `tabIndex` combination introduces regressions. The Radio input was already Tab-focusable before this change — the missing pieces were only the visible focus style and the Enter handler, so most of this can be simplified. Details inline.
> 
> Tests: none of the new behavior is covered — add cases for Enter selecting once, click selecting once, and loading items not being selectable.

Addressed comments, added test coverage.

src/components/common/SelectableCardList/components/SelectableCardListItem.test.tsx
```diff
@@ -94,4 +94,79 @@ describe('SelectableCardListItem', () => {
...
+
+    // get the card container and try to click it
+    const cardContent = container.querySelector('[class*="MuiCardContent"]')
+    if (cardContent) {
+      fireEvent.click(cardContent)
+    }
+
+    // should not call onSelect due to pointerEvents: 'none'
+    expect(onSelect).not.toHaveBeenCalled()
+  })
```
tomas562: This test passes vacuously: `[class*="MuiCardContent"]` never matches because `RcSesCard` renders children in a `Box`, not MUI `CardContent`, so the `if` guard skips the click and the assertion holds no matter what the component does. Dispatching the click wouldn't fix it either — jsdom doesn't enforce `pointer-events: none`, so `fireEvent.click` would reach the card's `onClick` even though a real browser blocks it. Assert the two mechanisms directly instead: no radio tab stop while loading (keyboard path) and the pointer-events style on the card root (mouse path).

```suggestion
  test('loading card is not selectable', () => {
    const onSelect = vi.fn()
    const { container } = renderItem(
      <SelectableCardListItem
        title='Loading'
        selected={false}
        onSelect={onSelect}
        isLoading
        listItems={[]}
      />,
    )

    // no radio is rendered while loading, so there is no keyboard tab stop
    expect(screen.queryByRole('radio')).toBeNull()

    // mouse clicks are blocked by pointer-events on the card root
    const card = container.querySelector('.MuiCard-root')
    expect(card).toHaveStyle('pointer-events: none')
  })
```

src/components/common/SelectableCardList/components/SelectableCardListItem.tsx
```diff
@@ -33,11 +33,26 @@ const SelectableCardListItem = ({
...
+          }
+        }}
         sx={{
           p: 0,
           height: '1.25rem',
           width: '1.25rem',
           margin: '0.125rem 0rem',
+          '&:focus-visible': {
+            outline: 'none',
+          },
```
tomas562: Dead code — keyboard focus lands on the hidden `<input>` inside the Radio, never on the root span this `&:focus-visible` targets, and MUI's default indicator here is the ripple, not an outline. The card-level `&:has(input:focus-visible)` border already provides the visible indicator.

```suggestion
```

src/components/common/SelectableCardList/components/SelectableCardListItem.tsx
```diff
@@ -78,8 +93,12 @@ const SelectableCardListItem = ({
         borderRadius: '0.75rem',
         padding: { xs: '1rem' },
         borderColor: selected ? palette.primary.main : palette.grey[300],
-        transition: 'all 0.2s ease',
+        transition: 'border-color 0.2s ease',
         gap: { xs: '0.5rem', md: '0.25rem' },
+        '&:has(input:focus-visible)': {
+          borderColor: palette.grey[900],
+          outline: 'none',
```
tomas562: No-op — the card is a non-focusable div, so it never has an outline to suppress; the `borderColor` line above is the whole effect.

```suggestion
```

tomas562 (commented):
All previous comments are addressed — the button/tabIndex removal, propagation stops, kebab-case transition, and `&:has(input:focus-visible)` all check out, and the new Enter/click tests cover the double-call regression. One remaining issue in the loading-state test plus two dead-code nits, suggestions inline.

tomas562 (approved):
All suggestions applied and CI is green — approving.

PR #109, opened 2026-07-01
SAV-6113: fix SearchInput issue

saukaite:
## 🔗 Task

https://jira.registrucentras.lt/jira/browse/SAV-6113

## 🧾 Summary

Added fixes, so that enter key works consistently across all environments (mobile + desktop), form pattern established for proper submission handling.

✅ ENTER press works on both desktop/mobile
✅ Search form is isolated with - `<Box component='form'>`

## 📸 Screenshots

Check in the Storybook (updated with the latest changes):

https://ses-react-storybook.registrucentras.lt/preview/SAV-6113-fix-search-input/

## ✅ Checklist

* [x] Tests added/updated

## ⚠️ Risks

N/A

### DIFF
--- a/src/components/form/inputs/SearchInput/index.tsx
+++ b/src/components/form/inputs/SearchInput/index.tsx
@@ -42,6 +42,14 @@ type Props = Pick<TControllerProps, ImmediateControllerProps> &
     }
   }
 
+/**
+ * Search input component with react-hook-form integration.
+ *
+ * ⚠️ Warning: This component renders an inner `<form>` element and uses `stopPropagation()`
+ * to prevent submit events from bubbling to parent forms. If nested inside another form,
+ * React will log a `validateDOMNesting` warning in development mode — this is expected
+ * and does not affect functionality.
+ */
 const RcSesSearchInput = React.forwardRef<HTMLInputElement, Props>((props, ref) => {
   const [hasSearchAttempted, setHasSearchAttempted] = React.useState(false)
   const internalInputRef = useRef<HTMLInputElement>(null)
@@ -117,15 +125,6 @@ const RcSesSearchInput = React.forwardRef<HTMLInputElement, Props>((props, ref) 
 
   const hasSearchValue = `${value ?? ''}`.trim().length > 0
   const visibleErrors = hasSearchAttempted ? errors : undefined
-
-  const handleSearchClick: NonNullable<ButtonProps['onClick']> = (event) => {
-    setHasSearchAttempted(true)
-    mergedSearchButtonProps.onClick?.(event)
-
-    if (!event.defaultPrevented && hasSearchValue) {
-      onSearch(value ?? '')
-    }
-  }
 
   const handleFormSubmit: React.FormEventHandler<HTMLFormElement> = (event) => {
     event.preventDefault()
@@ -259,8 +258,12 @@ const RcSesSearchInput = React.forwardRef<HTMLInputElement, Props>((props, ref) 
 
         {shouldRenderSearchButton && (
           <RcSesButton
+            type='submit'
             disabled={disabled || mergedSearchButtonProps.disabled || !hasSearchValue}
-            onClick={handleSearchClick}
+            onClick={(e) => {
+              mergedSearchButtonProps.onClick?.(e)
+              if (e.defaultPrevented) e.preventDefault()
+            }}
             sx={[
               {
                 flexShrink: 0,
--- a/src/components/form/inputs/SearchInput/index.tsx
+++ b/src/components/form/inputs/SearchInput/index.tsx
@@ -144,7 +144,12 @@ const RcSesSearchInput = React.forwardRef<HTMLInputElement, Props>((props, ref) 
 
     if (!event.defaultPrevented && event.key === 'Enter') {
       event.preventDefault()
-      internalInputRef.current?.form?.requestSubmit()
+      const form = internalInputRef.current?.form
+      if (form?.requestSubmit) {
+        form.requestSubmit()
+      } else {
+        form?.dispatchEvent(new Event('submit', { cancelable: true, bubbles: true }))
+      }
     }
   }
 
--- a/src/components/form/inputs/SearchInput/index.tsx
+++ b/src/components/form/inputs/SearchInput/index.tsx
@@ -28,7 +28,6 @@ type TSearchButtonProps = Partial<ButtonProps>
 type Props = Pick<TControllerProps, ImmediateControllerProps> &
   Pick<TWrapperProps, ImmediateWrapperProps> & {
     onSearch: (value: string) => void
-    onSubmit?: (value: string) => void
     required?: boolean
     onlyNumbers?: boolean
     placeholder?: TFieldProps['placeholder']
@@ -67,7 +66,6 @@ const RcSesSearchInput = React.forwardRef<HTMLInputElement, Props>((props, ref) 
     errors,
     label,
     onSearch,
-    onSubmit,
     required,
     onlyNumbers = false,
     placeholder,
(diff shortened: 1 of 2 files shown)


src/components/form/inputs/SearchInput/index.tsx
```diff
@@ -213,6 +217,8 @@ const RcSesSearchInput = React.forwardRef<HTMLInputElement, Props>((props, ref)
       {...mergedWrapperProps}
     >
       <Box
+        component='form'
```
tomas562, line 225: This component takes react-hook-form `control`, so consumers will typically render it inside their own `<form onSubmit={handleSubmit(...)}>`. That produces a nested `<form>`, which is invalid HTML (React logs a `validateDOMNesting` warning in dev), and the inner form's `submit` event bubbles: `handleFormSubmit` calls `preventDefault()` but not `stopPropagation()`, so Enter in the search field will also fire the ancestor form's `onSubmit` — RHF's `handleSubmit` ignores `defaultPrevented`, so the outer form would validate and submit. Add `event.stopPropagation()` in `handleFormSubmit`, plus a test that an ancestor form's `onSubmit` is not called.

src/components/form/inputs/SearchInput/index.tsx
```diff
@@ -126,22 +125,27 @@ const RcSesSearchInput = React.forwardRef<HTMLInputElement, Props>((props, ref)
...
+      onSubmit?.(value ?? '')
+      onSearch(value ?? '')
+    }
+  }
+
+  const handleFormSubmit: React.FormEventHandler<HTMLFormElement> = (event) => {
+    event.preventDefault()
+    setHasSearchAttempted(true)
+
+    if (hasSearchValue) {
```
tomas562, line 136: The removed Enter branch called `onBlur()` when the value was empty — that marked the RHF field as touched, so in `onTouched`/`onBlur` validation modes the consumer's `errors` got populated and the required error rendered after an empty search attempt. Now empty Enter only sets `hasSearchAttempted`, the field is never touched, and the error may never appear. Restore the `onBlur()` call in the empty branch here.

src/components/form/inputs/SearchInput/index.tsx
```diff
@@ -28,6 +28,7 @@ type TSearchButtonProps = Partial<ButtonProps>
 type Props = Pick<TControllerProps, ImmediateControllerProps> &
   Pick<TWrapperProps, ImmediateWrapperProps> & {
     onSearch: (value: string) => void
+    onSubmit?: (value: string) => void
```
tomas562: Every code path calls `onSubmit?.(value)` immediately followed by `onSearch(value)` with the same value — the two callbacks are indistinguishable, so the new prop adds API surface without adding a capability. Either give `onSubmit` distinct semantics (e.g. it also fires on empty value, or receives the form event) or drop it and keep `onSearch` as the single callback. If it stays, it should be documented in `SearchInput.stories.tsx`.

src/components/form/inputs/SearchInput/index.tsx
saukaite: Added `event.stopPropagation()` + test

src/components/form/inputs/SearchInput/index.tsx
saukaite: Restored `onBlur()`

src/components/form/inputs/SearchInput/index.tsx
saukaite: Removed `onSubmit` prop

src/components/form/inputs/SearchInput/index.tsx
```diff
@@ -130,18 +127,24 @@ const RcSesSearchInput = React.forwardRef<HTMLInputElement, Props>((props, ref)
...
-      setHasSearchAttempted(true)
+    if (!event.defaultPrevented && event.key === 'Enter') {
       event.preventDefault()
-
-      if (hasSearchValue) {
-        onSearch(value ?? '')
-      } else {
-        onBlur()
-      }
+      internalInputRef.current?.form?.requestSubmit()
```
tomas562: `requestSubmit()` is unavailable on Safari/iOS < 16, where Enter would silently do nothing - reintroducing SAV-6113 for exactly the mobile users this fixes. `form.submit()` is not a valid fallback (it skips the `onSubmit` handler and navigates); dispatching a cancelable submit event goes through React's handler.

```suggestion
      const form = internalInputRef.current?.form
      if (form?.requestSubmit) {
        form.requestSubmit()
      } else {
        form?.dispatchEvent(new Event('submit', { cancelable: true, bubbles: true }))
      }
```

src/components/form/inputs/SearchInput/index.tsx
```diff
@@ -130,18 +127,24 @@ const RcSesSearchInput = React.forwardRef<HTMLInputElement, Props>((props, ref)
...
+    event.preventDefault()
+    event.stopPropagation()
+    setHasSearchAttempted(true)
+
+    if (hasSearchValue) {
+      onSearch(value ?? '')
+    } else {
+      onBlur()
+    }
+  }
```
tomas562, line 141: `handleSearchClick` duplicates most of this handler and drifts from it: the form path calls `onBlur()` on an empty value to surface validation, the click path doesn't. Since the button now lives inside the form, `type='submit'` makes the form handler the single path:

```tsx
<RcSesButton
  type='submit'
  onClick={(e) => {
    mergedSearchButtonProps.onClick?.(e)
    if (e.defaultPrevented) e.preventDefault()
  }}
```

keeping only the `slotProps.searchButton.onClick` passthrough. (Suggestion block not possible here - the button JSX is outside the diff.)

tomas562 (commented):
Correct fix with the right shape - the native form pattern is what makes Enter uniform across environments, and the nested-form test is a good catch. Main asks: a requestSubmit fallback if the supported-browser matrix includes Safari/iOS < 16, and documenting the nested-form dev warning instead of "Risks: N/A". Button-as-type='submit' unification is a nice-to-have.

src/components/form/inputs/SearchInput/index.tsx
```diff
@@ -254,8 +263,12 @@ const RcSesSearchInput = React.forwardRef<HTMLInputElement, Props>((props, ref)
 
         {shouldRenderSearchButton && (
           <RcSesButton
+            type='submit'
             disabled={disabled || mergedSearchButtonProps.disabled || !hasSearchValue}
-            onClick={handleSearchClick}
+            onClick={(e) => {
+              mergedSearchButtonProps.onClick?.(e)
+              if (e.defaultPrevented) e.preventDefault()
+            }}
```
tomas562: `{...mergedSearchButtonProps}` is spread after `onClick`, so when a consumer passes `slotProps.searchButton.onClick` the spread replaces this wrapper entirely, and when no consumer handler exists the wrapper does nothing (`mergedSearchButtonProps.onClick?.(e)` is undefined). The `if (e.defaultPrevented) e.preventDefault()` line is a no-op either way — `defaultPrevented === true` means `preventDefault()` was already called. Native form semantics already provide the intended veto: a consumer `onClick` that calls `preventDefault()` blocks the submit on its own. Removing the wrapper keeps behavior identical.

```suggestion
```

src/components/form/inputs/SearchInput/index.tsx
```diff
@@ -42,6 +42,14 @@ type Props = Pick<TControllerProps, ImmediateControllerProps> &
     }
   }
 
+/**
+ * Search input component with react-hook-form integration.
+ *
+ * ⚠️ Warning: This component renders an inner `<form>` element and uses `stopPropagation()`
+ * to prevent submit events from bubbling to parent forms. If nested inside another form,
+ * React will log a `validateDOMNesting` warning in development mode — this is expected
+ * and does not affect functionality.
```
tomas562: "does not affect functionality" only holds for client-side rendering, where React builds the DOM via JS and the inner `<form>` actually exists. Under SSR/hydration the HTML parser drops the nested `<form>` tag, so `input.form` resolves to the parent form and Enter would submit it — the exact bug this PR fixes. Current consumers are SPAs, so this is a documentation-accuracy fix only.

```suggestion
 * React will log a `validateDOMNesting` warning in development mode — this is expected
 * for client-side rendering and does not affect functionality. Note this is not SSR-safe:
 * an HTML parser drops the nested `<form>` tag, so under SSR/hydration the Enter key
 * would submit the parent form instead.
```

tomas562:
LGTM

tomas562 (approved):

PR #110, opened 2026-07-15
SAV-6268: feat Create Switch component

TautvydasBliudzius:
## 🔗 Task

[](https://jira.registrucentras.lt/jira/browse/SAV-6268)

## 🧾 Summary

Added `RcSesSwitch` — a Toggle component wrapping MUI's `Switch`, styled per the DS spec (On/Off × Rest/Hover/Focus/Disabled).

- Controlled component (`checked` + `onChange` required)
- `role="switch"` + `aria-checked` added explicitly (MUI defaults to checkbox semantics)
- Enter key support added alongside native Space
- Requires either `label` or `ariaLabel` (enforced via TS types)

## 📸 Screenshots

image
image

## ✅ Checklist

* [+] Component added/updated in Storybook (if applicable)
* [+] Tests added/updated
* [+] UI matches approved design (Figma)
* [+] Accessibility reviewed (keyboard navigation, labels, semantics)


src/components/common/Switch/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+import palette from '@/theme/palette'
+
+type BaseProps = {
+  checked: boolean
+  disabled?: boolean
+  onChange: (event: React.ChangeEvent<HTMLInputElement>, checked: boolean) => void
+}
+
+export type RcSesSwitchProps = BaseProps &
+  ({ label: string; ariaLabel?: string } | { label?: undefined; ariaLabel: string })
```
tomas562: SAV-6268 „Būsenų logika #1“ requires the component to support both uncontrolled and controlled usage, but `checked`/`onChange` are required here, so the uncontrolled case is impossible. This also deviates from the library convention (`RcSesSimpleCheckbox` extends `MuiCheckboxProps` and forwards the rest), which drops `defaultChecked`, `name`, `inputRef` and `sx` — those are needed for react-hook-form integration and for the Advanced-list `Trailing=Toggle` composition planned as the next phase.

```suggestion
export type RcSesSwitchProps = SwitchProps &
  ({ label: string; ariaLabel?: string } | { label?: undefined; ariaLabel: string })
```

With this, import `SwitchProps` from `@mui/material`, destructure `{ label, ariaLabel, ...switchProps }` and spread `{...switchProps}` onto `<Switch>` — `defaultChecked` then covers the uncontrolled case with no extra code, and MUI's own `onChange` already has the `(event, checked)` signature.

src/components/common/Switch/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+const formControlLabelStyles = {
+  marginLeft: 0,
+  '& .MuiFormControlLabel-label': {
+    paddingLeft: '0.625rem',
+  },
+  '&.Mui-disabled .MuiFormControlLabel-label': {
+    color: palette.grey['500'],
+  },
+}
```
tomas562, line 17: SAV-6268 „Kraštiniai atvejai #1“ requires the switch to stay aligned with the **first line** of a wrapping label. `FormControlLabel` defaults to `align-items: center`, so with a two-line label the switch vertically centers across both lines. Align to the top and offset the 1.25rem switch to the label's first line — verify the exact offset against the DS type scale using the „Long label“ story:

```suggestion
const formControlLabelStyles = {
  marginLeft: 0,
  alignItems: 'flex-start',
  '& .MuiSwitch-root': {
    marginTop: '0.125rem',
  },
  '& .MuiFormControlLabel-label': {
    paddingLeft: '0.625rem',
  },
  '&.Mui-disabled .MuiFormControlLabel-label': {
    color: palette.grey['500'],
  },
}
```

src/components/common/Switch/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+      event.currentTarget.click()
+    }
+  }
+
+  const inputProps = {
+    'aria-label': ariaLabel ?? label,
+    role: 'switch',
+    'aria-checked': checked,
+    onKeyDown: handleKeyDown,
+  }
```
tomas562, line 33: Two issues here:

1. When a visible label is present, `aria-label` overrides it as the accessible name. With different texts this violates WCAG 2.5.3 (Label in Name) — the screen reader announces something other than what's on screen. The visible label already names the input via `FormControlLabel`, so `aria-label` should only apply in the label-less variant.
2. The explicit `aria-checked` is redundant — the native `checked` state of the underlying checkbox input maps to the switch state automatically, even with `role="switch"`.

```suggestion
  const inputProps = {
    ...(label === undefined && { 'aria-label': ariaLabel }),
    role: 'switch',
    onKeyDown: handleKeyDown,
  }
```

Test updates needed alongside: the „should prefer an explicit ariaLabel over the visible label text“ test asserts the opposite of this, and the `aria-checked` attribute assertions become `toBeChecked()` / `not.toBeChecked()`.

src/components/common/Switch/index.test.tsx
```diff
@@ -0,0 +1,190 @@
...
+      expect(input).toHaveAttribute('aria-checked', 'true')
+    })
+  })
+
+  describe('Accessible name', () => {
+    it('should associate the visible label with the switch via FormControlLabel', () => {
+      render(<RcSesSwitch label='Notifications' checked={false} onChange={vi.fn()} />)
+
+      expect(screen.getByRole('switch', { name: 'Notifications' })).toBeInTheDocument()
+    })
```
tomas562, line 40: SAV-6268 „Prieinamumas #2“ — clicking the visible label toggles the switch — is the main behavior `FormControlLabel` is here for, and it has no test. Add:

```tsx
it('should toggle when the visible label is clicked', () => {
  const handleChange = vi.fn()
  render(<RcSesSwitch label='Notifications' checked={false} onChange={handleChange} />)

  fireEvent.click(screen.getByText('Notifications'))

  expect(handleChange).toHaveBeenCalledTimes(1)
  expect(handleChange.mock.calls[0][1]).toBe(true)
})
```

tomas562 (commented):
Review against SAV-6268. Two acceptance-criteria gaps (uncontrolled usage, first-line label alignment), one accessibility issue (`aria-label` overriding the visible label), and a missing label-click test — details inline with suggestions.

tomas562 (approved):
LGTM

PR #115, opened 2026-07-24
SAV 6273 Advanced-list-item-v2

TautvydasBliudzius:
## 🔗 Task

[
](https://jira.registrucentras.lt/jira/browse/SAV-6273)

## 🧾 Summary

Adds a new AdvancedList / AdvancedListItem component pair — a single universal list-row component (Figma: Advanced-list-item-v2) that replaces having separate components per use case. It supports configurable leading accessories (checkbox, radio, reorder, drag, icon, avatar, index) and trailing accessories (actions, toggle, status badge, price, chevron, stepper), row selection/state (rest/selected/disabled/error), and an expandable content slot (GRM-12030 Phase 5) for inline forms/detail panels.

## 📸 Screenshots

image
image
image
image

## ✅ Checklist

[+] Component added/updated in Storybook (if applicable)
[+] Tests added/updated
[+] UI matches approved design (Figma)
[+] Accessibility reviewed (keyboard navigation, labels, semantics)

## ⚠️ Risks

Radio leading/trailing types currently render MUI's raw Radio directly instead of a RcSes-prefixed wrapper (unlike checkbox, switch, badge, button, which all have one) — there's no RcSesRadio component yet. It's flagged in-code (// change to RcSesRadio when it is implemented) but not addressed in this PR; worth a follow-up ticket so radio styling/behavior stays consistent with the rest of the design system.


src/components/common/SelectableCardList/components/SelectableCardListItem.tsx
```diff
@@ -97,6 +104,33 @@ const SelectableCardListItem = ({
         },
       }}
       title={titleTemplate}
+      headerAction={
+        isLoading ? (
+          <Skeleton variant='rounded' width={100} height={38} />
```
tomas562, line 110: One thing to fix before merge, in the unrelated `SelectableCardList` change: the `headerAction` skeleton is gated on `isLoading` only, so every loading card renders a 100x38 button skeleton even when `hasActionButton` is false. `SelectableCardList.renderSkeletons()` never passes it, so existing consumers get a phantom button in their loading state. Gating the whole expression on `hasActionButton` fixes it.

tomas562:
mobile view looks bad with more content:
image
image

TautvydasBliudzius:
> mobile view looks bad with more content: image image

Matau figmoj jau atsirado mobile version dizainas, tai dar skirsiu laiko sutvarkymui pagal jį

src/library/index.ts
```diff
@@ -99,7 +101,6 @@ export { RcSesSearchInput }
 export { RcSesSearchableField }
 export { RcSesSelect }
 export { RcSesTab, RcSesTabPanel, RcSesTabs, RcSesTabsWrapper }
-export { RcSesTooltip }
 export { RcSesTextField }
```
tomas562, line 103: The head merge commit deletes work that is already on `develop` rather than keeping both sides: the whole SAV-6270 Tooltip feature (component, test, story, `src/theme/light/MuiTooltip.ts` and its registration, and the `RcSesTooltip` / `RcSesTooltipProps` exports here and in `types.ts`), plus the SAV-6446 wrapping fix in `ListWithIconsItem`. `RcSesTooltip` would vanish from the next published build and break consumers that import it. No conflict is shown because `a455be2` merged those commits into the branch first and `09dd552` removed the files afterwards, which git reads as an intentional newer state.

This PR does not otherwise touch the barrels or the theme index, so restoring develop's version of the affected paths is exact:

```bash
git checkout origin/develop -- \
  src/components/common/Tooltip \
  src/stories/components/display/Tooltip.stories.tsx \
  src/theme/light/MuiTooltip.ts \
  src/theme/light/index.tsx \
  src/library/index.ts \
  src/library/types.ts \
  src/components/common/ListWithIcons/components/ListWithIconsItem.tsx
```

Then `git diff --stat origin/develop...HEAD` should list only `AdvancedListItem/index.tsx`, `SelectableCardListItem.tsx` and the two i18n files.

src/components/common/ListWithIcons/components/ListWithIconsItem.tsx
```diff
@@ -21,14 +21,14 @@ const ListWithIconsItem = ({
...
+        display: 'inline-flex',
+        alignItems: 'center',
         gap: 0.5,
-        maxWidth: '100%',
+        maxWidth: { xs: '100%', md: 'none' },
       }}
     >
       {Icon && (
-        <Box sx={{ display: 'flex', alignItems: 'center', height: 20 }}>
+        <Box sx={{ display: 'flex', alignItems: 'center' }}>
```
tomas562, line 31: SAV-6446 is reverted here. The `git checkout` in the comment above restores this file wholesale; if you prefer to fix it in place, this is the content part:

```suggestion
        display: 'flex',
        alignItems: 'flex-start',
        gap: 0.5,
        maxWidth: '100%',
      }}
    >
      {Icon && (
        <Box sx={{ display: 'flex', alignItems: 'center', height: 20 }}>
```

tomas562 (commented):
Round 2 review. Blocking: the head merge commit `09dd552` resolves the `develop` merge by deleting work already on `develop` (378 lines, 10 files) instead of keeping both sides:

- the whole SAV-6270 Tooltip feature: component, test, story, `src/theme/light/MuiTooltip.ts` and its registration, and the public `RcSesTooltip` / `RcSesTooltipProps` exports. The `common.Tooltip.trigger` i18n keys stay behind with no consumer. Merging as-is drops `RcSesTooltip` from the next published build.
- the SAV-6446 wrapping fix in `ListWithIconsItem` (commented inline).

Redo the merge keeping both, so this branch carries only the AdvancedListItem/SelectableCardListItem changes. Also note `a455be2` leaves `en.json` / `lt.json` invalid JSON, which is what pulled the Tooltip sources out during the hand-fix.

Remaining comments are on the new responsive AdvancedListItem layout.

src/components/common/ListWithIcons/components/ListWithIconsItem.tsx
```diff
@@ -38,8 +38,7 @@ const ListWithIconsItem = ({
           fontSize: '14px',
           lineHeight: '20px',
           fontWeight: 300,
-          whiteSpace: 'normal',
-          wordBreak: 'break-word',
+          whiteSpace: 'nowrap',
```
tomas562, line 41:
```suggestion
          whiteSpace: 'normal',
          wordBreak: 'break-word',
```
With `nowrap` a long meta value renders as one unbroken line past the card border on xs, where the `ul` is a column and each `li` is 100% wide.

src/components/common/ListWithIcons/components/ListWithIconsItem.tsx
```diff
@@ -82,7 +81,6 @@ const ListWithIconsItem = ({
         display: 'flex',
         alignItems: 'center',
         width: { xs: '100%', md: isLoading ? '120px' : 'auto' },
```
tomas562, line 83:
```suggestion
        width: { xs: '100%', md: isLoading ? '120px' : 'auto' },
        maxWidth: '100%',
```
Without this the `li` can exceed its container even once the text wraps.

src/components/common/SelectableCardList/components/SelectableCardListItem.tsx
```diff
@@ -97,6 +104,32 @@ const SelectableCardListItem = ({
...
+            }}
+            onClick={(e) => {
+              e.stopPropagation()
+              onActionButtonClick?.()
+            }}
+            variant='outlined'
+          >
+            {actionButtonLabel}
+          </RcSesButton>
+        ))
```
tomas562, line 131: `renderSkeletons` passes `isLoading` without `hasActionButton`, so gating the whole thing on `hasActionButton` makes the skeleton branch unreachable and brings back the header reflow it was preventing. `isLoading` needs to stay the outer condition:

```suggestion
        isLoading ? (
          <Skeleton variant='rounded' width={100} height={38} />
        ) : (
          hasActionButton && (
            <RcSesButton
              color='grey'
              sx={{
                height: 'auto',
                padding: '0.5rem 1rem',
                fontSize: '0.875rem',
                lineHeight: '1.25rem',
                '&.MuiButton-colorGrey': {
                  color: palette.grey[800],
                },
              }}
              onClick={(e) => {
                e.stopPropagation()
                onActionButtonClick?.()
              }}
              variant='outlined'
            >
              {actionButtonLabel}
            </RcSesButton>
          )
        )
```

src/components/common/AdvancedList/components/AdvancedListItem/index.tsx
```diff
@@ -0,0 +1,307 @@
...
+    return {
+      ...leading,
+      onChange: (checked: boolean) => {
+        leading.onChange(checked)
+        onExpandedChange?.(checked)
+      },
+    }
+  }, [canAttachExpandToLeading, leading, onExpandedChange])
+
+  useEffect(() => {
```
tomas562, line 102: This effect only ever focuses `selectorRef`, and `selectorRef` is attached solely when `canAttachExpandToLeading` is true (leading `radio` / `checkbox`). For `leading: 'none' | 'reorder' | 'icon'` with `expandedContent`, an external collapse leaves focus inside the panel that is being marked `inert`, so `activeElement` drops to `body` and the next Tab restarts at the top of the page. Add a fallback focus target (the row container with `tabIndex={-1}`, or the expand toggle) for those leading types.

The deleted comment above this effect explained why it has to be an effect on `expanded` rather than part of the leading control's `onChange`; that reasoning is still load-bearing, keep it.

src/components/common/AdvancedList/components/AdvancedListItem/index.tsx
```diff
@@ -0,0 +1,307 @@
...
+          : undefined,
+        '&:focus-visible, &:focus-within': !isDisabled
+          ? { borderColor: palette.primary.main, boxShadow: focusRingShadow }
+          : undefined,
+      }}
+    >
+      <Stack
+        direction='row'
+        alignItems='flex-start'
+        flexWrap={{ xs: 'wrap', sm: 'nowrap' }}
```
tomas562, line 194: This row switches at `sm` (600px) but the `ListWithIcons` nested in it switches at `md` (900px), so 600-899px gets both branches at once: the header is in its `nowrap` single-line desktop layout while the meta items still stack one per row at 100% width, stretching the row to roughly 3x its intended height with the trailing control centred against it. Align the two on one breakpoint.

The expanded-panel indent at line 289 (`marginLeft: leading.type !== 'none' ? '2rem' : 0`) was also left non-responsive, so on xs the panel is indented 2rem against a leading control that is no longer in a left column.

src/components/common/AdvancedList/components/AdvancedListItem/index.tsx
```diff
@@ -0,0 +1,307 @@
...
+        )}
+
+        <Box
+          sx={{
+            display: 'flex',
+            flexDirection: 'row',
+            alignItems: 'flex-start',
+            gap: '0.75rem',
+            minWidth: 0,
+            flex: { xs: '1 1 100%', sm: '1 1 0%' },
```
tomas562, line 217: `flex: { xs: '1 1 100%' }` wraps the content column onto its own line for every xs render, including a plain checkbox item with a short title and no trailing slot: the checkbox ends up alone on line 1 with the label below it, doubling item height and breaking the control/label association. Gate the full-width basis on there being a trailing slot.

src/components/common/AdvancedList/components/AdvancedListItem/index.tsx
```diff
@@ -0,0 +1,307 @@
...
+              <ListWithIcons items={metaItems} layout='horizontal' />
+            )}
+          </Box>
+        </Box>
+
+        {trailing.type !== 'none' && (
+          <Box
+            sx={{
+              alignSelf: 'center',
+              order: { xs: -1, sm: 0 },
```
tomas562, line 270: `order: -1` plus `marginLeft: 'auto'` on xs puts the trailing control above the title. With `leading: 'none'` it is the only element on the first line, so a stepper or price floats alone on its own row with an empty gap to its left and the title below it. Visual order also stops matching DOM order (the title comes first in the DOM), so keyboard and screen-reader order disagree with what is on screen. Only hoist the trailing box when there is a leading control to share that line with.

tomas562 (approved):
Looks good now

PR #117, opened 2026-07-27
SAV-5935: InlineAlert component

jsnaras:
## 🔗 Task

SAV-5935

## 🧾 Summary

New InlineAlert component with 5 variants. Based on container queries for optimal UX on mobile and in small parent containers on desktop

## 📸 Screenshots

inline-alert-1

inline-alert-2

inline-alert-3

## ✅ Checklist

* [x] Tests added/updated

## ⚠️ Risks

None identified

### DIFF
--- a/src/components/common/InlineAlert/index.test.tsx
+++ b/src/components/common/InlineAlert/index.test.tsx
@@ -1,0 +1,125 @@
+import { ThemeProvider } from '@mui/material/styles'
+import { fireEvent, render, screen } from '@testing-library/react'
+import { vi } from 'vitest'
+
+import theme from '@/theme/light'
+
+import RcSesInlineAlert from './index'
+
+vi.mock('react-i18next', () => ({
+  useTranslation: () => ({
+    t: (key: string) => key,
+  }),
+}))
+
+const renderAlert = (ui: React.ReactElement) =>
+  render(<ThemeProvider theme={theme}>{ui}</ThemeProvider>)
+
+describe('RcSesInlineAlert', () => {
+  test('renders children', () => {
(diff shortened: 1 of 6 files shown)


tomas562:
Nice work — the container-query layout and the per-severity `role`/`aria-live` mapping are both genuine improvements over what we have today.

One thing we need to resolve: we already have `RcSesAlert`, and it sits very close to this component. What was the driver behind building a new one from scratch? The only requirement I can see that `RcSesAlert` can't express is CTA + close together (MUI only renders its close button when `action == null`) — was that it, or was there more?

I don't see both living in this library when they do the same job. Maybe `RcSesAlert` could be deprecated in a future version, but I need to understand the reasoning first.

jsnaras:
Good point, agree on unifying them instead of having two. As an external dev, I wanted to avoid any breaking changes to existing RcSesAlert, per spec/design the alert now:
- needs to show both CTA and close button (most important reason)
- needs `role="status"` per severity (MUI hardcodes `role="alert"`)
- needs "neutral" severity
- could benefit from container-query

I think we could do the breaking change here: rename `RcSesInlineAlert` to `RcSesAlert` and rename `variant` propt to `severity` (leaves `variant` free for visual options). Not sure how to handle the "container" full-page banner - technically it's a banner, not an inline alert?

Or leave both and just deprecate RcSesAlert? Please let me know what you prefer

tomas562:
Thanks for the context. I'm waiting on confirmation from the designer about the CTA + close button combination - it looks like a genuine edge case to me rather than the default. I'll come back to this PR once all the information is collected.

tomas562:
Closing this PR in favour of https://github.com/registrucentras/rc-ses-react-components/pull/129. We decided to go with single Alert component.

PR #120, opened 2026-08-04
SAV-6447: Select component to wrap long text and improve input ha…

JenLoVangas:
…ndling

## 🔗 Task

[SAV-6447](https://jira.registrucentras.lt/jira/browse/SAV-6447)

## 🧾 Summary

Fixes a mobile UI issue in the JAR extract ordering flow where long extract title/goal text was truncated.
Updated styles so the text wraps to the next line instead of being cut off, improving readability

## 📸 Screenshots

image
image

## ✅ Checklist

* [x] Tests added/updated


src/components/form/inputs/Select/index.tsx
```diff
@@ -104,6 +103,15 @@ function RcSesSelect(props: Props) {
     return options.find((o) => o.value === value) ?? null
   }, [value, options, multiple])
 
+  const selectedSingleLabel =
+    !multiple && resolvedValue && !Array.isArray(resolvedValue) ? resolvedValue.label : ''
+  let resolvedInputValue: string | undefined
+  if (multiple) {
+    resolvedInputValue = dropdownSearch ? undefined : inputValue
+  } else {
+    resolvedInputValue = ''
```
tomas562: Forcing `inputValue` to `''` for single selects leaves the underlying input empty, so its `placeholder` now paints next to the selected value and gets clipped by MUI's `text-overflow: ellipsis` on `.MuiAutocomplete-input`. The result is a stray greyed `P...` beside the wrapped label. Reproduced in Storybook: a single select with a value and a `placeholder` shows it, the same select without a `placeholder` does not, and it affects the `dropdownSearch` path too. Before this change `onInputChange` set `inputValue` to the selected label, so the input was non-empty and the placeholder stayed hidden. Pass the placeholder only when nothing is selected, for example `placeholder={selectedSingleLabel ? undefined : placeholder}` on the `TextField` at line 257.

image

src/components/form/inputs/Select/index.tsx
```diff
@@ -186,7 +194,7 @@ function RcSesSelect(props: Props) {
         open={open}
         onOpen={() => setOpen(true)}
         multiple={multiple}
-        limitTags={multiple ? limitTags : undefined}
+        limitTags={multiple ? -1 : undefined}
```
tomas562: `limitTags` is still declared in the public props at line 33 but is no longer destructured, so this hardcoded `-1` makes it a dead prop. Verified: rendering with `limitTags={1}` and two selected values renders both chips. Consumers can pass it, TypeScript accepts it, and nothing happens. Either honour it as the collapse threshold or drop it from the props type and call the removal out as a breaking change. Nothing in the workspace passes it today, so this is latent rather than an active break.

src/components/form/inputs/Select/index.tsx
```diff
@@ -260,34 +267,62 @@ function RcSesSelect(props: Props) {
...
+              startAdornment:
+                !multiple && selectedSingleLabel ? (
+                  <Box className='rc-ses-select-single-value'>{selectedSingleLabel}</Box>
+                ) : (
+                  params.InputProps.startAdornment
+                ),
               inputProps: {
                 ...params.inputProps,
-                readOnly: dropdownSearch,
+                readOnly: dropdownSearch || !multiple,
```
tomas562: With the input `readOnly` and `inputValue` pinned to `''`, the combobox no longer exposes its value to assistive tech. Verified: with an option selected the input's `value` is `""`, and the label lives in a `pointer-events: none` div in the `startAdornment`, so screen readers announce the field label and an empty combobox. Keyboard interaction itself is intact, confirmed by test: Tab focuses the combobox, ArrowDown opens the listbox, ArrowDown plus Enter sets the form value, and Clear resets it to `null`. Giving the value div an id and referencing it from the input via `aria-describedby`, or folding the selected label into the input's `aria-label`, closes the gap. Worth noting the `readOnly` change itself is defensible: `filterOptions` at line 151 already returned every option unless `dropdownSearch` was set, so typing never filtered anything.

src/components/form/inputs/Select/index.tsx
```diff
@@ -260,34 +267,62 @@ function RcSesSelect(props: Props) {
                 ? {
                     '& .MuiAutocomplete-inputRoot': {
                       alignItems: 'center',
-                      pb: '0 !important',
-                      pt: '0 !important',
+                      pb: '.25rem !important',
+                      pr: '3.5rem !important',
```
tomas562: The end adornment width is now encoded in three places with `!important`: `pr: '3.5rem'` here, `marginRight: '2rem'` on `.MuiAutocomplete-input` at lines 290 and 309, and `maxWidth: 'calc(100% - 3.5rem)'` on `.MuiAutocomplete-tag` at line 287. They will drift apart if the adornment changes. A single shared constant for the reserved width keeps them in sync.

src/theme/light/MuiInputBase.ts
```diff
@@ -14,7 +14,9 @@ const MuiInputBase: Components['MuiInputBase'] = {
         fontSize: '.9375rem',
         height: '1.125rem',
         lineHeight: '1.125rem',
+        overflowWrap: 'anywhere',
```
tomas562: `overflowWrap` and `wordBreak` on the `.MuiInputBase-input` slot have no effect on single-line `<input>` elements, which cannot wrap. They do apply to `multiline`, but MUI puts `MuiInputBase-input` on the textarea as well, so the new `.MuiInputBase-inputMultiline` block at line 81 already duplicates them. One of the two should go, and since this is a global theme file it lands on every input in the library.

src/theme/light/MuiAutocomplete.ts
```diff
@@ -82,6 +82,9 @@ const MuiAutocomplete: Components['MuiAutocomplete'] = {
 
           '.rc-ses-select-option-label': {
             display: 'block',
+            overflowWrap: 'anywhere',
+            whiteSpace: 'normal',
+            wordBreak: 'break-word',
```
tomas562: `wordBreak: 'break-word'` is deprecated and its effect is exactly `overflow-wrap: anywhere` plus `word-break: normal`, so it adds nothing next to the `overflowWrap: 'anywhere'` on the line above. `whiteSpace: 'normal'` is also just the initial value. The same trio appears in five places across this file and the Select component, so dropping the redundant two keeps the theme readable. Minor: `wordBreak` at line 99 also lands after `fontWeight`, breaking the alphabetical ordering the rest of the block follows.

src/stories/components/inputs/Select.stories.tsx
```diff
@@ -90,7 +90,10 @@ function SelectDemo(args: any) {
           disabled={disabled}
           slotProps={slotProps}
           options={[
-            { value: 'pasirinkimas-1', label: 'Pasirinkimas #1' },
+            {
+              value: 'pasirinkimas-1',
+              label: 'Labai labai ilgas patikrinimui skirtas Pasirinkimas #1 ',
```
tomas562: The long test label replaced the default in all seven `SelectDemo` blocks and in the generated code snippet at line 228, trailing space included, so every state preview and the copyable example now carry it. CONTRIBUTING asks for short, user-oriented story args that reflect real usage. One dedicated long-label story keeps the defaults clean and documents the fix better. Happy to hand over a story that covers the empty and preselected single select, the `dropdownSearch` variant and the multiple select with long values, which is what I used to verify this PR.

tomas562:
**Tests**

The checklist marks `Tests added/updated` as done, but the diff contains no test files and there is no Select test anywhere in the repo - the component directory holds only `hooks`, `index.tsx` and `Select.types.ts`. This change alters `readOnly`, `inputValue`, tag rendering and the public prop surface, so a test that a single select renders its full label and stays selectable by keyboard would be worth having. Other form inputs already have a working harness to copy from, for example `src/components/form/inputs/PhoneInputFormControl/index.test.tsx`.

**Multi-select values are still truncated**

Only the single-select path actually wraps. In the multiple case `.MuiAutocomplete-tag` gets `maxWidth: 'calc(100% - 3.5rem)'` and MUI's Chip ellipsizes its own label, so a long value renders as `Juridinių asmenų registro išpl...`. That is normal chip behaviour and both fields in SAV-6447 look like single selects, so this is probably fine. Worth confirming no field in the JAR flow is a multiple select before treating the ticket as fully covered.

tomas562:
Before select was searchable
image
now I cant type in select's input
image

I found a consumer in ses-ui 
image
it uses this functionality so it's a breaking change now

src/components/common/SelectableCardList/components/SelectableCardListItem.tsx
```diff
@@ -82,19 +79,21 @@ const SelectableCardListItem = ({
   )
   return (
     <RcSesCard
+      aria-label={`Select ${title}`}
+      aria-pressed={selected}
+      role='button'
+      tabIndex={isLoading ? -1 : 0}
       onClick={onSelect}
+      onKeyDown={isLoading ? undefined : handleKeyDown}
```
tomas562: This card is not part of SAV-6447. The ticket is about long select values being clipped, and the JAR extract flow does not depend on how a selectable card handles keyboard input. CONTRIBUTING asks a pull request to solve one task, so this belongs in its own change.

The new semantics are also wrong. The card becomes `role='button'` with `aria-pressed`, but it still contains a real `<input type="radio">` carrying the same `Select ${title}` accessible name, so the same item is announced twice as two different widgets, and an interactive descendant inside `role="button"` is invalid ARIA. This is a single-choice list, so the correct model is `role="radio"` with `aria-checked` inside a `role="radiogroup"` container. That container is `SelectableCardList/index.tsx`, which this PR does not touch, so the correct model cannot land here anyway.

Restoring the previous implementation keeps the native radio as both the accessible control and the keyboard target. Verified: with the five suggestions on this file applied the result matches `develop`, `npx eslint src/components/common/SelectableCardList` is clean and all 17 tests in that folder pass.

```suggestion
      onClick={onSelect}
```

src/components/common/SelectableCardList/components/SelectableCardListItem.tsx
```diff
@@ -82,19 +79,21 @@ const SelectableCardListItem = ({
...
       sx={{
         cursor: isLoading ? 'default' : 'pointer',
         pointerEvents: isLoading ? 'none' : 'auto',
         border: '0.125rem solid',
         borderRadius: '0.75rem',
         padding: { xs: '1rem' },
         borderColor: selected ? palette.primary.main : palette.grey[300],
-        transition: 'border-color 0.2s ease',
+        transition: 'all 0.2s ease',
         gap: { xs: '0.5rem', md: '0.25rem' },
```
tomas562: The focus indicator goes back with it. `&:has(input:focus-visible)` was the only designed focus treatment on this card, and without it the card falls back to the browser default outline, which CONTRIBUTING rules out under "Išsaugokite aiškiai matomą fokusavimo elgseną". `transition: 'all'` also animates every animatable property on a card that only ever changes its border colour.

```suggestion
        transition: 'border-color 0.2s ease',
        gap: { xs: '0.5rem', md: '0.25rem' },
        '&:has(input:focus-visible)': {
          borderColor: palette.grey[900],
        },
```

src/components/common/SelectableCardList/components/SelectableCardListItem.tsx
```diff
@@ -33,18 +41,7 @@ const SelectableCardListItem = ({
...
-        }}
-        onClick={(e) => e.stopPropagation()}
-        onKeyDown={(e) => {
-          if (e.key === 'Enter') {
-            e.preventDefault()
-            e.stopPropagation()
-            onSelect()
-          }
-        }}
+        tabIndex={-1}
```
tomas562: `tabIndex={-1}` takes the radio out of the tab order while leaving it in the accessibility tree, so it is still announced but can no longer be reached or operated. Restoring the handlers makes the radio the control again and keeps a click on it from reaching the card handler twice.

```suggestion
        checked={selected}
        onChange={(e) => {
          e.stopPropagation()
          onSelect()
        }}
        onClick={(e) => e.stopPropagation()}
        onKeyDown={(e) => {
          if (e.key === 'Enter') {
            e.preventDefault()
            e.stopPropagation()
            onSelect()
          }
        }}
```

src/components/common/SelectableCardList/components/SelectableCardListItem.tsx
```diff
@@ -23,6 +24,13 @@ const SelectableCardListItem = ({
   onSelect,
   isLoading = false,
 }: Props) => {
+  const handleKeyDown = (event: KeyboardEvent<HTMLDivElement>) => {
+    if (event.key === 'Enter' || event.key === ' ') {
+      event.preventDefault()
+      onSelect()
+    }
+  }
+
```
tomas562: `handleKeyDown` is unused once the card stops being a button.

```suggestion
}: Props) => {
```

src/components/common/SelectableCardList/components/SelectableCardListItem.tsx
```diff
@@ -1,4 +1,5 @@
 import { Box, Radio, Skeleton, Stack, Typography } from '@mui/material'
+import { KeyboardEvent } from 'react'
```
tomas562: `KeyboardEvent` is unused after that.

```suggestion
import { Box, Radio, Skeleton, Stack, Typography } from '@mui/material'
```

src/components/form/inputs/Select/index.tsx
```diff
@@ -141,22 +159,36 @@ function RcSesSelect(props: Props) {
...
 
-    const filteredValues = new Set(filteredOptions.map((o) => o.value))
-    const filtered = realOptions.filter((opt) => filteredValues.has(opt.value))
+      return [searchOption, ...filtered]
+    }
+
+    if (dropdownSearch && hasCustomGroupBy) return opts
 
-    return [searchOption, ...filtered]
+    if (!inputValue.trim()) return opts
```
tomas562: A single select can no longer be changed once it holds a value. This branch filters by the component's own `inputValue` state, and MUI resets `inputValue` to the selected option's label whenever the value changes, including on mount with a preselected value. The filter then collapses the list to the option that is already selected. MUI's built-in filter guards against exactly this by substituting `''` when `inputValue` equals the selected label, but this custom `filterOptions` ignores the `state` argument, so the guard never runs. `closeDropdownWithClear` does not help, because MUI's reset fires after the close.

Verified against both branches with a single select over three options:

| Scenario | develop | this branch |
| --- | --- | --- |
| Preselected `option-1`, open dropdown | 3 options | 1 option |
| Select `3rd option`, close, reopen | 3 options | 1 option |
| Nothing selected, type `2nd` | no filtering | filters correctly |

This is the SAV-6447 flow itself: choose the extract and the goal, then try to change the choice and the remaining options are gone. `npm run lint`, all 204 tests and CI are green, so nothing currently catches it.

Setting `readOnly` on the input while a value is selected does not fix it, because MUI resets `inputValue` regardless of `readOnly`. Guarding the filter does. With the suggestion applied both dropdown scenarios offer all three options again, switching from `option-1` to `3rd option` works, type ahead filtering with nothing selected still works, and the two new tests in `index.test.tsx` still pass. A regression test that reopens a select which already holds a value would keep this from coming back.

```suggestion
    if (!inputValue.trim() || inputValue === selectedSingleLabel) return opts
```

src/components/form/inputs/Select/index.tsx
```diff
@@ -104,6 +105,20 @@ function RcSesSelect(props: Props) {
...
+  const selectedSingleLabel =
+    !multiple && resolvedValue && !Array.isArray(resolvedValue) ? resolvedValue.label : ''
+
+  const selectedTagCount =
+    multiple && Array.isArray(resolvedValue) ? resolvedValue.length : 0
+  const hasLimitTagCounter =
+    multiple && (limitTags ?? -1) > -1 && selectedTagCount > (limitTags ?? -1)
+  const reservedMultipleEndAdornmentWidth = hasLimitTagCounter
+    ? `${reservedSingleEndAdornmentWidth * 1.5}`
+    : `${reservedSingleEndAdornmentWidth * 1.1}`
```
tomas562: The reserved end adornment width is derived from `reservedSingleEndAdornmentWidth` through `* 1.5` and `* 1.1` multipliers and returned as a string, which is then interpolated back into `calc(...rem)` at lines 294, 297 and 520. The two factors carry no meaning at the call sites and will be guessed at the next time the adornment changes. Declaring the two widths directly as numbers, next to `reservedSingleEndAdornmentWidth`, keeps the intent readable and drops the string round trip.

The chip max width is also still expressed twice: globally on `.MuiChip-root` at line 297 with `!important`, and per chip in `sx` at line 520. One of the two is enough, and the per chip `sx` is the narrower place for it.

src/stories/components/inputs/Select.stories.tsx
```diff
@@ -88,9 +88,13 @@ function SelectDemo(args: any) {
...
           label={label}
           disabled={disabled}
+          placeholder={placeholder}
           slotProps={slotProps}
           options={[
-            { value: 'pasirinkimas-1', label: 'Pasirinkimas #1' },
+            {
+              value: 'pasirinkimas-1',
+              label: 'Pasirinkimas #1 ',
+            },
```
tomas562: `'Pasirinkimas #1 '` carries a trailing space. It is repeated in all seven `SelectDemo` blocks and in the generated code snippet at line 236, which is what library users copy out of Storybook. This looks like residue from the long label that was reverted, so restoring the original one line option drops both the trailing space and the reformatting noise.

```suggestion
            { value: 'pasirinkimas-1', label: 'Pasirinkimas #1' },
```

PR #123, opened 2026-08-12
Sav 6270 select controller slot props

tomas562:
## 🔗 Task

https://jira.registrucentras.lt/browse/SAV-6270

## 🧾 Summary

`RcSesSelect` hardcoded `shouldUnregister: true`, so a consumer could not keep the
value when the field unmounts. This blocks the AdvancedList radio branches, where
switching a radio variant unmounts the fields of the branch being left and the form
loses the data.

`RcSesSearchInput` and `RcSesSearchableField` already allow this through
`slotProps.controller`. Select now does the same, with the default kept at
`shouldUnregister: true`, so nothing changes for existing consumers.

While checking the other form inputs, `RcSesRadioButtonGroup`, `RcSesNumberStepper`,
`RcSesFileUpload` and `RcSesFileDropzone` turned out to declare `controller` in their
public `slotProps` type without ever spreading it into `useController`, so the prop
type-checked and did nothing. They now spread it too.

Also added a Storybook story under `components/display/AdvancedList` that runs the
radio branches through react-hook-form, with a `shouldUnregister` toggle and panels
showing the form state next to the branch-filtered submit payload.

## ✅ Checklist

* [ x] Tests added/updated
* [ x] UI matches approved design (Figma)
* [ x] Accessibility reviewed (keyboard navigation, labels, semantics)

## ⚠️ Risks

LOW

### DIFF
--- a/src/components/form/inputs/Select/index.test.tsx
+++ b/src/components/form/inputs/Select/index.test.tsx
@@ -1,5 +1,6 @@
 import { ThemeProvider } from '@mui/material/styles'
 import { fireEvent, render, screen, waitFor } from '@testing-library/react'
+import { useState } from 'react'
 import { useForm } from 'react-hook-form'
 
 import theme from '@/theme/light'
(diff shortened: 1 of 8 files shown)


src/components/form/inputs/Select/index.test.tsx
```diff
@@ -99,4 +138,22 @@ describe('RcSesSelect', () => {
 
     expect(screen.getByRole('combobox')).toHaveAttribute('readonly')
   })
+
+  test('drops the value from the form by default when the field unmounts', () => {
```
saukaite, line 142: [nit-pick] small inconsistency, but in the same test file both LT & EN languages are used for the test names.

saukaite (approved):
LGTM ✅ checked new stories in the Storybook as well

src/components/form/inputs/Select/index.test.tsx
tomas562: Yes, also noticed this. but probably it's a task for other PR. We should keep everything in English and tighten contribution rules.

PR #125, opened 2026-08-13
docs(storybook): document the validation-error pattern

tomas562:
## Why

A consumer team hit what looked like a bug: passing `fieldState.error?.message` into `SelectableCardList`'s `error` prop made every card disappear, leaving only the message.

That is the prop working as designed. `error` is a data-loading failure state - the node it receives is rendered instead of the cards and the pagination (`SelectableCardList/index.tsx`, `error != null ? error : ...`), and the unit test asserts the list is hidden. It has behaved this way since the component was added.

The problem is discoverability. The only error-shaped example in Storybook was the `Error` story, which looks like a generic error slot, and `FormControlWrapper` had no `errors` example at all - so nothing anywhere showed the correct validation pattern.

## What changed

Stories only, no component or behaviour changes.

**SelectableCardList**
- new `ValidationError` story: the list wrapped in `RcSesFormControlWrapper` with `errors`, cards stay visible and selectable, message renders below
- `Error` story keeps its name (existing Storybook links stay valid) and gains a description stating `items` is populated yet still replaced, and that this is a data-loading state
- component description now spells out `error` vs the wrapper's `errors` and points at the two stories

**FormControlWrapper**
- new `WithError` story, plus a note that the pattern applies to any wrapped control including non-inputs, and that a translated fallback is shown when `errors.type` is `required` with no `message`

### DIFF
--- a/src/stories/components/display/SelectableCardList.stories.tsx
+++ b/src/stories/components/display/SelectableCardList.stories.tsx
@@ -9,6 +9,7 @@ import SelectableCardList, {
   type SelectableCardListItemData,
   type SelectableCardListProps,
 } from '@/components/common/SelectableCardList'
+import RcSesFormControlWrapper from '@/components/form/components/FormControlWrapper'
 
 const getListItems = (index: number): ListWithIconsItemData[] => {
   switch (index) {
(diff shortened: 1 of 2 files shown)


tomas562:
Storybook preview deployed from this branch:

https://ses-react-storybook.registrucentras.lt/preview/docs/selectable-card-list-error-stories/

PR #126, opened 2026-08-13
SAV-6509: add CardV2 component

saukaite:
## 🔗 Task

https://jira.registrucentras.lt/jira/browse/SAV-6509

## ✅ Checklist

* [ ] Tests added/updated


src/examples/CardMultipleStepForm/index.tsx
```diff
@@ -1,80 +0,0 @@
-import { Button, Typography } from '@mui/material'
```
saukaite, line 1: Removed this file as its an example only and uses removed **RcSesCardFormContainer** component which will be replaced with the **StepCard** (follow-up PR).

tomas562 (commented):
The shell architecture matches what was agreed on SAV-6509: ReactNode slots, `variant`, `theme`, `fullHeight`, a separate `RcSesCardHeader` with `headingLevel`, the badge inside the heading's accessible name, a decorative icon. Dropping the nested-heading DOM in `SelectableCardListItem`, where an `<h5>` wrapped a `Radio` and an `<h3>`, is a real accessibility improvement, and moving that component off `RcSesCard` is the right call. Verified that no consumer in the workspace (ses-ui, mfe-host, mfe-navigation, bdar-report-ui, demo-providers) imports `RcSesCard`, `RcSesCardFormContainer`, `RcSesImageCard` or `SelectableCardList`, so the breaking API change costs nothing today.

Five inline comments cover the hardcoded ImageCard surface, the four boxed surfaces now in the library, the missing migration entries, the footer losing mobile stacking, and hardcoded typography.

### Header block: three gaps, all found by walking the approved designs

1. **`icon` has no component and no story.** Nothing in `Card.stories.tsx` passes it, and the Icon-tile it is meant to hold (SAV-6476, the rounded square in the Group-card and Category-tile designs) does not exist - the library only has the circular `IconWithCircularBackground`. So the one slot the acceptance criteria name explicitly, decorative Icon-tile with `aria-hidden`, is both unreachable and unverified.
2. **No trailing-affordance slot.** Category-tile (SAV-6477) is specified as "header ON su rodykle". That arrow is decorative, must be `aria-hidden`, and must not become a second tab stop next to the stretched link. Passing it through `actions` does render, since `actions` is an unvalidated `ReactNode`, but that prop is documented as up to 2 CTAs.
3. **No alignment control.** `CardHeader.tsx:27` pins `alignItems: 'flex-start'`, so the icon sits against the title's first line. In the Category-tile design it is centered against a two-line text block, and `RcSesCardHeader` accepts no `sx`, so a composition cannot correct it.

`actions` also has no story of its own, so two of the six header slots ship with no visual coverage.

### Token checks against Figma, while the values are being settled

- Radius is `0.75rem` here, where the previous `MuiCard` theme used `1rem`.
- The `brand` theme pairs `primary[50]` fill with a `primary[200]` border that the Group-card design does not obviously show.
- Card padding is deliberately heavier at the bottom, which is right with a footer and will read off-balance on a header-only tile like Category-tile.

### Smaller items

1. Six baseline files should not be in the commit. `components-display-cardv2--{card,header-off,subcard-inside-card,themes,with-badge}.png` are leftovers from the pre-rename CardV2 story, and `components-form-cardformcontainer--horizontal-demo.png` was modified rather than deleted although its story file is gone. 11 baselines added for 6 stories. Playwright does not fail on orphaned snapshots, so CI stays green either way.
2. The exported `RcSesCardProps` no longer describes the component. Its real prop type is the unexported `ShellProps`, so a consumer typing a props object as `RcSesCardProps` cannot include `sx` or `component`. Export the resolved type.
3. `SelectableCardListItem.test.tsx:217` asserts on `container.querySelector('.MuiBox-root')`, which matches every Box in the subtree and passes only because the root happens to be first. Use the root test id.
4. `docs/SAV-5648-backlog.md:340-342` is now stale: after this PR no `RcSes*` wrapper renders `MuiCard` either, so it joins `MuiCardHeader` and `MuiCardContent` in the table above it.
5. `RcSesCardFormContainer` is removed while SAV-6544 is still in progress, and the SAV-6509 agreement had the wizard staying as `RcSesCardFormContainer` rebuilt internally on the new shell. Nothing consumes it, so this is a release-content decision rather than a defect: 2.0.0 would ship with neither the old container nor Step-card.

saukaite:
Closing in favor of: https://github.com/registrucentras/rc-ses-react-components/pull/132

